PR #5, opened 2018-03-06
T/4:  Minimal table support

### LINKED ISSUE #4: Minimal table support
I'd like to propose incremental work on table feature as it will take some time and effort to fully develop even in MVP form.

As the minimal table support I'd like to propose:
- Table conversion (basic upcast/downcast that will not destroy existing data) with:
    - `tbody`/`thead`/`tfoot` parsing,
    - `tr`,`td`,`th` handling,
    - attributes parsing (at least `colspan` & `rowspan`).
- Insert table button (probably temporary) that will insert 2x2 table.
- Insert table command (maybe with parameters already like rows & columns).

And basically that. No cursor support, no better UI, no widget and probably buggy in one way or another but I think that setting small steps along the way might be useful and probably better to review during that time rather one big PR.

jodator:

Feature: Initial table support. Closes #4.

---

As in it's task description this PR provides very, very, very basic `<table>` support:
- converts `<table>` to `table` model,
- stores number of heading rows & columns as `table` attributes,
- converts `rowspan` and `colspan` attributes,
- `InsertTableCommand` that can insert `rows x columns` table,
- `insertTable` UI button that inserts a 2x2 table (shouldn't be this `table:insert` :thinking:)

It does not have:
- widget,
- selection (ie it creates "magic paragraph" like bug if you try to click after a table),
- UI/UX.

As such I'm quite happy with this as you can somehow edit an existing table.


jodator:
I've found at least one flaw with the conversion that should be fixed.

Issue #1, opened 2017-11-02
Table UI and UX


oleq:
This is the follow-up of the discussion in https://github.com/ckeditor/ckeditor5/issues/610. The features that need to be implemented to make the plugin(s) working are described as [Milestone 1 (MVP)](https://github.com/ckeditor/ckeditor5/issues/610#issuecomment-337267343).

There are various factors to consider in the design:
1. The table insertion should happen using the visual "grid" in the main editor toolbar.
   1. There should be a way to insert a huge table too, e.g. 120x300, using some inputs. The inputs could be the UI elements used as current grid dimension labels (requires research) or a separate UI.
1. The UI manipulating table cells/rows/styles should be in the context of the table. Probably a balloon attached to the cell in which the user selection is anchored.
   1. Note that attaching a balloon to the entire table will cause issues when the table is very long (see next point) or when other editor balloons (like global toolbar) are in use.
1. The UI must work when the table has 100+ rows and neither its first or last row is visible in the viewport.
1. The UI should probably disappear as the user types to uncover other cells and their content and show up again as the user stops typing.
1. Table editing tools should not use balloons to avoid a situation when a balloon is attached to another balloon. It becomes a problem when the balloon "arrows" overlap for certain arrangements. We should use dropdowns instead.
1. The table header feature should allow conversion of first N-rows of the table into a header. Users usually create content and leave the formatting as the last step.
1. The cross–selection between the cells should be blocked in MVP.

dkonopka:
Proposals below. Icons still in progress, but would be great if you say your opinion about them (especially merge icon).

![ck5-ui-table](https://user-images.githubusercontent.com/10219857/32329425-5c248276-bfdd-11e7-8727-023efd3ce59e.png)

fredck:
The designs look good.

I'm just wondering about the drop-down button.  Shouldn't we have icons there? After all, these toolbars are, in theory, configurable, so one could include any of the drop-down buttons into the toolbar itself. Just like described in https://github.com/ckeditor/ckeditor5-ui/issues/333.

In the other hand, we can also agree that this is yet another kind of button, which simply shows the list of options as text, "period". For the specific case of row/column tools, I'm more than fine about that.

fredck:
As for the merge icons, I like both options, especially the arrows because they look different than the rest (otherwise all icons look the same and you get confused).

[...]

PR #23, opened 2018-05-29
T/1: Implement Table UI

jodator:

Feature: Implement the base Table UI.

---

### Additional information

Requires changes in 
* ui: https://github.com/ckeditor/ckeditor5-ui/pull/390
* theme lark: https://github.com/ckeditor/ckeditor5-theme-lark/pull/171


oleq:
Why is the table inserted after current block instead of splitting it (see v4 https://ckeditor.com/ckeditor-4/)? It's not very intuitive IMO.

kapture 2018-05-30 at 13 29 45

oleq:
There's somethign wrong with the position of the `BalloonToolbar` when used for the link in the table because it remains attached to the table, not to the link

screen shot 2018-05-30 at 13 30 07

Not sure which piece of code is broken, `TableUI` or something else.

jodator:
> Why is the table inserted after current block instead of splitting it

That's why we need this Table UI fast :D Things as this will come as one can actually test tables with UI better. It's the way how `InsertTableCmmand` was implemented. As this is not UI related I'll report this as a follow up and fix it in this iteration.

jodator:
> Not sure which piece of code is broken, TableUI or something else.

After quick checkup with I there might be two issues:
1. the table feature will show & update position of the balloon but somehow the link plugin contents are inserted so either race condition of which plugin should show baloon on mouse click
2. when updating selection with keyboard the link plugin updates position of the balloon with data that looks OK (`{target: a.ck-link_selected}`) but the balloon stays attached to the table.

So I can see that table toolbar could better check when to show it's toolbar.

ps.: I've found some glitch here:
https://github.com/ckeditor/ckeditor5-ui/blob/9b4d88863f2f817a42dd436a3b185a3849ca124a/src/panel/balloon/contextualballoon.js#L196

The position returned here does not match the last one passed to the updatePosition as in the `_stack` the returned position is for the table and not for the link. But without further investigation I'm not able to see if it is problem of how table uses toolbar or the link.

oleq:
There's a problem with the color of the arrow too

screen shot 2018-05-30 at 14 38 28

oleq:
> the table feature will show & update position of the balloon but somehow the link plugin contents are inserted so either race condition of which plugin should show baloon on mouse click

It could boil down to https://github.com/ckeditor/ckeditor5/issues/852 and this is bad :/

oleq:
Table toolbar also hijacks the main balloon editor toolbar

image

oleq:
BTW where's the drag handler as in https://github.com/ckeditor/ckeditor5-table/issues/1#issuecomment-392034926? Is this also a followup?

jodator:
> BTW where's the drag handler as in #1 (comment)? Is this also a followup?

Hmm. Good question. I might assumed this as a widget feature (so also for an image) and as such could have follow-up. If not R- this PR and I'll add this. WDYT?

oleq:
> Hmm. Good question. I might assumed this as a widget feature (so also for an image) and as such could have follow-up. If not R- this PR and I'll add this. WDYT?

OK, I created an [issue](https://github.com/ckeditor/ckeditor5-widget/issues/40) in ckeditor5-widget. By default (until we implement the propoer drag and drop), only the table widget is supposed to have this selection (later: drag) handler. So it must be explicitely enabled by a widget (e.g. as a property).

oleq:
We also need to figure out how to keep the focused cell border on the top

image

Note: `position: relative` and `z-index` don't help.

jodator:
@oleq there's also this issue:

selection_057

dtwist:
Use outline instead of border:
https://codepen.io/david-twist/pen/NMQOwO?editors=1000

```css
.highlight {
  outline: 2px solid rgb(71, 164, 245);
  box-shadow: 0px 2px 6px rgba(0, 0, 0, .2) inset;
}
```

src/tabletoolbar.js
```diff
@@ -0,0 +1,194 @@
...
+	get _isVisible() {
+		return this._balloon.visibleView == this._toolbar;
+	}
+}
+
+/**
+ * Items to be placed in the table toolbar.
+ * This option is used by the {@link module:table/tabletoolbar~TableToolbar} feature.
+ *
+ * Assuming that you use the {@link module:table/tableui~TableUI} feature below toolbar items will be available
```
oleq: `below toolbar items/the following toolbar items`

src/tabletoolbar.js
```diff
@@ -0,0 +1,194 @@
...
+ *		const tableConfig = {
+ *			toolbar: [ 'tableRow', 'tableColumn', 'mergeCell', 'splitCell' ]
+ *		};
+ *
+ * Of course, the same buttons can also be used in the
+ * {@link module:core/editor/editorconfig~EditorConfig#toolbar main editor toolbar}.
+ *
+ * Read more about configuring toolbar in {@link module:core/editor/editorconfig~EditorConfig#toolbar}.
+ *
+ * @member {Array.<String>} module:table~TableConfig#toolbar
```
oleq: Where's the @interface TableConfig defined BTW because I couldn't find it? I can't find these docs after building under `api/module_core_editor_editorconfig-EditorConfig.html`

src/tableui.js
```diff
@@ -8,14 +8,26 @@
...
 /**
- * The table UI plugin.
+ * The table UI plugin. It introduces:
+ *
+ * * The `'insertTable'` dropdown,
+ * * The `'tableColumn'` dropdown,
+ * * The `'tableRow'` dropdown,
+ * * The `'mergeCell'` dropdown.
+ *
+ * The `'tableColumn'`, `'tableRow'`, `'mergeCell'` are best to be used in conjunction with {@link module:table/tabletoolbar~TableToolbar}.
```
oleq: `are best to be used in conjunction with/work best with` ;-)

src/ui/inserttableview.js
```diff
@@ -0,0 +1,181 @@
...
+		} );
+
+		// Add grid boxes to table selection view.
+		for ( let index = 0; index < 100; index++ ) {
+			const boxView = new TableSizeGridBoxView();
+
+			// Listen to box view 'over' event which indicates that mouse is over this box.
+			boxView.on( 'over', () => {
+				// Translate box index to the row & column index.
+				const row = parseInt( index / 10 );
```
oleq: Please use `Math.floor`. It's confusing why `parseInt` is used in this place.

src/ui/inserttableview.js
```diff
@@ -0,0 +1,181 @@
...
+			// Grid box is highlighted when its row & column index belongs to selected number of rows & columns.
+			const isOn = itemRow < rows && itemColumn < columns;
+
+			boxView.set( 'isOn', isOn );
+		} );
+	}
+}
+
+/**
+ * Single grid box view element.
```
oleq: "A single". A reference to the class using the box would also be nice ;-)

src/ui/utils.js
```diff
@@ -0,0 +1,51 @@
...
+ * @param {module:core/editor/editor~Editor} editor The editor instance.
+ */
+export function repositionContextualBalloon( editor ) {
+	const balloon = editor.plugins.get( 'ContextualBalloon' );
+
+	balloon.updatePosition( getBalloonPositionData( editor ) );
+}
+
+/**
+ * Returns the positioning options that control the geometry of the
```
oleq: that controls

src/utils.js
```diff
@@ -0,0 +1,51 @@
...
+ * @returns {module:engine/view/element~Element}
+ */
+export function toTableWidget( viewElement, writer ) {
+	writer.setCustomProperty( tableSymbol, true, viewElement );
+
+	return toWidget( viewElement, writer );
+}
+
+/**
+ * Checks if a given view element is an table widget.
```
oleq: a table

src/utils.js
```diff
@@ -0,0 +1,51 @@
...
+ * @module table/utils
+ */
+
+import { toWidget, isWidget } from '@ckeditor/ckeditor5-widget/src/utils';
+import { getParentTable } from './commands/utils';
+
+const tableSymbol = Symbol( 'isImage' );
+
+/**
+ * Converts a given {@link module:engine/view/element~Element} to an table widget:
```
oleq: a table

src/utils.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @param {module:engine/view/element~Element} viewElement
+ * @returns {Boolean}
+ */
+export function isTableWidget( viewElement ) {
+	return !!viewElement.getCustomProperty( tableSymbol ) && isWidget( viewElement );
+}
+
+/**
+ * Checks if an table widget is the only selected element.
```
oleq: a table

oleq:
I also see a warning in the manual test

```
toolbarview-item-unavailable: The requested toolbar item is unavailable. Read more: https://docs.ckeditor.com/ckeditor5/latest/framework/guides/support/error-codes.html#error-toolbarview-item-unavailable
{name: "splitCell"}
 ``

File: src/commands/setheadercolumncommand.js [hunk 10]
```diff
@@ -0,0 +1,78 @@
...
+		const model = this.editor.model;
+		const doc = model.document;
+		const selection = doc.selection;
+
+		const position = selection.getFirstPosition();
+		const tableParent = getParentTable( position );
+
+		this.isEnabled = !!tableParent;
+
+		this.value = this.isEnabled && this._isInHeading( position.parent, tableParent );
```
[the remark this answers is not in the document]
jodator: I'll update this piece of code as it does not make sense to check if selection is in heading when there is no table.

src/tabletoolbar.js
```diff
@@ -0,0 +1,194 @@
...
+import { isTableWidgetSelected } from './utils';
+import { repositionContextualBalloon, getBalloonPositionData } from './ui/utils';
+
+const balloonClassName = 'ck-toolbar-container';
+
+/**
+ * The table toolbar class. Creates an table toolbar that shows up when the table widget is selected.
+ *
+ * Toolbar components are created using the editor {@link module:ui/componentfactory~ComponentFactory ComponentFactory}
+ * based on the {@link module:core/editor/editor~Editor#config configuration} stored under `table.toolbar`.
```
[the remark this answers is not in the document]
jodator: Yep it is missing - my bad.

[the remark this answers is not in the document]
jodator: OTOH it's copied from image toolbar - so it is described as there ;)

The @interface was created anyway as that was missing..

jodator:
@oleq All the issues were fixed - the only one that I didn't touch [is this one](https://github.com/ckeditor/ckeditor5-table/pull/23#discussion_r192709428) as I wrote there I've wrote this docs basically the same way as `ImageToolbar` one (hence an table in so many places ;) ).

PR #33, opened 2018-06-07
t/28: The BalloonToolbar should be enabled for a table content

### LINKED ISSUE #28: Toolbar positioning and collision issues
There are 2 serious issues regarding the table toolbar:

## Collision with the balloon editor

image

Effectively there's no way for format the content in the table when using the balloon toolbar.

## Wrong position when using other tools

Other features using the `ContextualBalloon` should not use the position of the table toolbar (the link balloon should be attached to the link in the cell).

image

----

A follow-up of 
* https://github.com/ckeditor/ckeditor5-table/pull/23#issuecomment-393127885
* https://github.com/ckeditor/ckeditor5-table/pull/23#issuecomment-393147292

oleq:
Fix: The `BalloonToolbar` should be enabled for a table content. Link editing should be possible in the table. Closes #28.

----

**Note:** Requires https://github.com/ckeditor/ckeditor5-ui/pull/406 to fix the issue with the link balloon attached to the table instead of a link in the cell.

### DIFF
--- a/src/tabletoolbar.js
+++ b/src/tabletoolbar.js
@@ -10,7 +10,7 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import ToolbarView from '@ckeditor/ckeditor5-ui/src/toolbar/toolbarview';
 import ContextualBalloon from '@ckeditor/ckeditor5-ui/src/panel/balloon/contextualballoon';
-import { isTableWidgetSelected } from './utils';
+import { isTableWidgetSelected, isTableContentSelected } from './utils';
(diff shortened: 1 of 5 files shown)


Mgsy:
There is one case I've found. It occurs in Safari and only if you use a touchpad and make a 'soft-click' (quick tap). 

**Steps to reproduce**
1. Create new table and fill it with some text.
1. Select some text and add a new link.
1. Put the caret outside of the link.
1. 'Soft-click' on the link.

**Current result**
The selection is inside the link but the balloon doesn't show up.

**GIF**
bug_cke5

**Notes**
If you'll `hard-click` on the link, the balloon will appear.

PR #45, opened 2018-06-12
t/2: The MVP of table content styles

### LINKED ISSUE #2: Define default table styles
We'll need to define the default style for tables and, once we'll implement the "table style" feature, a couple of useful alternatives.

The https://alistapart.com/article/web-typography-tables article covers many aspects of table's typography and should help us defining the defaults.

oleq:

Feature: The MVP of table content styles. Moved styles related to the editing to ckeditor5-theme-lark (see #2).

---

### Additional information

* Requires: https://github.com/ckeditor/ckeditor5-theme-lark/pull/176

### About

**Note:** It's an MVP which means we may still improve the styles in the future (doesn't close #2).

* Centered layout similar to default image styles,
* By default the table is **not** full-width because it looks odd when it's 2x2 and spans 1000px
* `min-width` of the table cell + `padding` to make an empty table look more user–friendly; easier to place the caret inside,
* Bold text in header cells to stress their meaning,
* A very slight background of header cells to avoid situations when the body cells have bold text and there's no obvious boundary between the header and the content (also when there's no content at all),
* A very slight background of a selected cell which works with the default nested editable's border  (blue) to help users navigate and understand the current position in big tables (since the toolbar is attached to the table, not to the cell).

screen shot 2018-06-12 at 10 29 54

screen shot 2018-06-12 at 10 28 53

### DIFF
--- a/theme/table.css
+++ b/theme/table.css
@@ -3,24 +3,30 @@
  * For licensing, see LICENSE.md.
  */
 
-.ck table.ck-widget th.ck-editor__nested-editable {
-	border-color: inherit;
-	border-style: inset;
-}
+.ck.ck-content table {
+	/* Give the table some air and center it horizontally */
+	margin: 1em auto;
 
-.ck table.ck-widget td.ck-editor__nested-editable {
-	border-color: inherit;
-	border-style: inset;
-}
+	/* The table cells should have slight borders */
+	border-collapse: collapse;
+	border-spacing: 0;
 
-.ck table.ck-widget td.ck-editor__nested-editable.ck-editor__nested-editable_focused,
-.ck table.ck-widget td.ck-editor__nested-editable:focus {
-	background-color: inherit;
-	color: inherit;
-}
+	/* The outer border of the table should be slightly darker than the inner lines. */
+	&:not(:hover) {
+		outline: 1px solid hsl(0, 0%, 70%);
+		outline-offset: -1px;
+	}
 
-.ck table.ck-widget th.ck-editor__nested-editable.ck-editor__nested-editable_focused,
-.ck table.ck-widget th.ck-editor__nested-editable:focus {
-	background-color: inherit;
-	color: inherit;
+	& td,
+	& th {
+		min-width: 2em;
+		padding: .4em;
+		text-align: center;
+		border-color: hsl(0, 0%, 85%);
+	}
+
+	& th {
+		font-weight: bold;
+		background: hsl(0, 0%, 98%);
+	}
 }
--- a/src/table.js
+++ b/src/table.js
@@ -13,6 +13,8 @@ import TableEditing from './tableediting';
 import TableUI from './tableui';
 import Widget from '@ckeditor/ckeditor5-widget/src/widget';
 
+import '../theme/table.css';
+
 /**
  * The table plugin.
  *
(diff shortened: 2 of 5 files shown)


Reinmar:
I see R+es from @jodator and @dkonopka. There's still https://github.com/ckeditor/ckeditor5-theme-lark/pull/176 to be reviewed. @dkonopka, could you check it and merge both of them when done?

Reinmar:
I think that you even reported a ticket for it to discuss it (and I ignored it back then :ashamed:). Let's be consistent with how other features do this for now.

dkonopka:
@oleq If we are choosing similar styles to the image/widget, shouldn't we change current focus border color (yellow) of table and change it to the blue?

tablefocus

widget-focus

theme/table.css
```diff
@@ -3,24 +3,30 @@
...
-}
+	/* The outer border of the table should be slightly darker than the inner lines. */
+	&:not(:hover) {
+		outline: 1px solid hsl(0, 0%, 70%);
+		outline-offset: -1px;
+	}
+
+	& td,
+	& th {
+		min-width: 2em;
```
dkonopka: IMHO it should be possible to easily increase/decrease the minimum width of a cell via CSS variable. 
Same thing with padding, but it is not much important like width I guess. WDYT?

Reinmar: It's hard to say how many people will want to change it. I'd say that at this stage we don't need this var yet. Maybe we'll understand more once we get some feedback.

Reinmar:
I forgot about one thing – I was wondering whether the table should be centred by default? 

I wanted to say that no because it looks weird when it's small. And I guess it's kinda true.

At the same time, a small image will be centred:

image

But is an image the same case? Small images will not be a frequent scenario. While narrow tables could be.

So, WDYT?

oleq:
> @oleq If we are choosing similar styles to the image/widget, shouldn't we change current focus border color (yellow) of table and change it to the blue?

This is actually interesting. I thought those are default widgets styles and I took them for granted assuming they will work the same in both cases. Can you check that out?

> But is an image the same case? Small images will not be a frequent scenario. While narrow tables could be.

IMO this is a different case and personally, I'd expect this kind of content to be centered by default. Also, consider the case where a table is 80% width of the editable (due to its content) – when aligned to the left, it will look strange or even buggy. If we center it, it's alright.

dkonopka:
> This is actually interesting. I thought those are default widgets styles and I took them for granted assuming they will work the same in both cases. Can you check that out?

It all boils down to missing `ck-widget_selected` class in the `table.ck-widget`.

> I'd expect this kind of content to be centered by default. 

☝️ I'm for centered too. 

Anyway, I'm sure it will be a lot of requests about "alignment" feature of table, no matter if we align it by default to the left or center 😄 .

Reinmar:
> Also, consider the case where a table is 80% width of the editable (due to its content) – when aligned to the left, it will look strange or even buggy. If we center it, it's alright.

Good point. 👍

oleq:
> It all boils down to missing ck-widget_selected class in the table.ck-widget.

I think it's alright. When you focus the caption of the image, then it acts just like the table. Bot image captions and table cells are nested editables. The widget is not selected in this situation.

dkonopka:
> I think it's alright. When you focus the caption of the image, then it acts just like the table. Both image captions and table cells are nested editables. The widget is not selected in this situation.

So there would be never a situation with `ck-widget` selected or it will be introduced with whole table selection?

jodator:
> So there would be never a situation with ck-widget selected or it will be introduced with whole table selection?

This is how whole table is selected:
selection_011

PR #48, opened 2018-06-12
Fix: The SetHeaderRowCommand should only analyze rows that are in hea…

### LINKED ISSUE #35: Table layout gets broken when setting header rows
jun-08-2018 11-32-58

1. Open http://localhost:8125/ckeditor5-table/tests/manual/table.html
2. Make the first row a normal row.
3. Make it a header row.

The layout breaks.

jodator:

Fix: The `SetHeaderRowCommand` should only analyze rows that are in heading section after expanding header. Closes #35.

---

### Additional information

* Another bug introduced by wrong analysis scope - one row too much was analyzed when command was checking for cells that might expand beyond a heading.

### DIFF
--- a/src/commands/setheaderrowcommand.js
+++ b/src/commands/setheaderrowcommand.js
@@ -111,8 +111,10 @@ function getOverlappingCells( table, headingRowsToSet, currentHeadingRows ) {
 	const cellsToSplit = [];
 
 	const startAnalysisRow = headingRowsToSet > currentHeadingRows ? currentHeadingRows : 0;
+	// We're analyzing only when headingRowsToSet > 0.
+	const endAnalysisRow = headingRowsToSet - 1;
 
-	const tableWalker = new TableWalker( table, { startRow: startAnalysisRow, endRow: headingRowsToSet } );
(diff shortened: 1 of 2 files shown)


oleq:
There's still something wrong with the algorithm

kapture 2018-06-13 at 16 09 20

jodator:
> There's still something wrong with the algorithm
@oleq: The header cells must be split before setting header rows because they cannot span over `<thead>` to `<tbody>. In the gif above:
1. first set header row command breaks that table cell into two cells so all the cells from `<thead>` section doesn't span over.
2. the second execute does the same - the cell in heading section has `rowspan=2` and must be broken down if heading is moved in the middle of it.

jodator:
Such table is wrong:
```html
<table>
	<thead>
		<tr>
			<th rowspan=2>Wrong! - will be rendeder as rowspan=1</th>
			<th></th>
		</tr>
	</thead>
	<tbody>
		<tr>
			<td>this will be rendered as a row without one cell</td>
		</tr>
	</tbody>
</table>
```

PR #59, opened 2018-06-15
T/51: Use <figure> for table widget

### LINKED ISSUE #51: The output HTML should use <figure>
The output should match the format proposed by http://ckeditor.github.io/editor-recommendations/features/table.html

jodator:

Fix: The output HTML should use `<figure>`. Closes #51.

---

### Additional information

* Also changes in theme lark: https://github.com/ckeditor/ckeditor5-theme-lark/tree/t/ckeditor5-table/51.

### DIFF
--- a/tests/_utils/utils.js
+++ b/tests/_utils/utils.js
@@ -3,6 +3,79 @@
  * For licensing, see LICENSE.md.
  */
 
+/**
+ * @param {Number} columns
+ * @param {Array.<String>} tableData
+ * @param {Object} [attributes]
+ *
+ * @returns {String}
+ */
+export function modelTable( tableData, attributes ) {
+	const tableRows = makeRows( tableData, 'tableCell', 'tableRow', 'tableCell' );
+
+	return `<table${ formatAttributes( attributes ) }>${ tableRows }</table>`;
+}
+
+/**
+ * @param {Number} columns
+ * @param {Array.<String>} tableData
+ * @param {Object} [attributes]
+ *
+ * @returns {String}
+ */
+export function viewTable( tableData, attributes = {} ) {
+	const headingRows = attributes.headingRows || 0;
+
+	const thead = headingRows > 0 ? `<thead>${ makeRows( tableData.slice( 0, headingRows ), 'th', 'tr' ) }</thead>` : '';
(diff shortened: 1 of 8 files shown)


Reinmar:
Just to be sure – does the feature still upcast a bare `<table>` (not wrapped)?

theme/table.css
```diff
@@ -3,28 +3,31 @@
  * For licensing, see LICENSE.md.
  */
 
-.ck.ck-content table {
-	/* Give the table some air and center it horizontally */
+.ck.ck-content figure.table {
```
oleq: 002n is enough according to our standards (e.g. `.ck-content .image {}`) so `.ck-content .table` should be OK.

jodator:
> Just to be sure – does the feature still upcast a bare `<table>` (not wrapped)?

yep - it's in the manual tests - only one table is wrapped in `<figure>` and the other ones are left as is.

theme/table.css
oleq: In fact, because of that, the table styles override the generic nested editable styles and we've lost the blue border around it when the cell is focused.

image

tests/_utils/utils.js
```diff
@@ -93,7 +48,8 @@ export function formatTable( tableString ) {
 		.replace( /<\/thead>/g, '\n</thead>' )
 		.replace( /<\/tbody>/g, '\n</tbody>' )
 		.replace( /<\/tr>/g, '\n</tr>' )
-		.replace( /<\/table>/g, '\n</table>' );
+		.replace( /<\/table>/g, '\n</table>' )
+		.replace( /<\/figure>/g, '\n</table>' );
```
dtwist: Is this a typo? (replacing "figure" with "table")

Reinmar: Nice catch :D Thanks.

@jodator, I changed it and all tests passed. Are you sure this is used and needed?

jodator: @Reinmar this is a helper method to nicely print out a table contents - it is used on both expected and actual results thus this slipped unnoticed...

This method helps a lot when some test fails and you read sinon output to easily spot what's wrong:

with `formatTable()`:

selection_043

without:

selection_044

I've also gradually updated other tests to use those helper methods as writing expected model table string is tedious....

Issue #47, opened 2018-06-12
Merging cells right / left might provide to broken layout


jodator:
![screencast 2018-06-12 14_43_02](https://user-images.githubusercontent.com/247363/41291099-0de4b826-6e4f-11e8-9e4e-6ccfeba4860b.gif)

The above actions will produce such table:

```html
<table>
    <tr>
        <td>a</td>
        <td colspan="2">bc</td>
    </tr>
    <tr>
        <td colspan="2">ab</td>
        <td>c</td>
    </tr>
```
which is rendered as:

![selection_006](https://user-images.githubusercontent.com/247363/41291229-6118e530-6e4f-11e8-9185-32472f01f043.png)

I don't really know it this issue is important or easy to fix.

Reinmar:
This is funny. I guess that the "merge left" option should not be enabled. It allows breaking the table layout so it should be disabled.

jodator:
> This is funny. I guess that the "merge left" option should not be enabled. It allows breaking the table layout so it should be disabled.

Yeah... but the tricky part is how to detect such case?

Reinmar:
You know the layout of the table. So you can also predict what applying this command will cause to it. Am I right?

jodator:
> You know the layout of the table. So you can also predict what applying this command will cause to it. Am I right?

Right now it seems to me that you'd have to check some special case of missing column (the table should have 3 columns) but this one has 2 cells in each row (as I'm writing it - maybe it is not such complex case after all...) and with rows it is straight forward as table cells are child of tableRow element. For columns we have to guess...

jodator:
ps.: similar thing happens in CKEditor 4

PR #60, opened 2018-06-15
Fix: Remove empty row in MergeCellCommand.

### LINKED ISSUE #16: Merging an only cell in a row results in empty row
When merging a cell that is only child of a row results in empty row leftover:

screencast 2018-05-24 16_24_54

Either should be fixed by the mergin algorithm (easy check as it is done in other places) or by a postfixer. 
Probably fixing it in `MergeCellCommand` is a way to go.

jodator:

Fix: The `MergeCellCommand` should check if merging cells results in empty row. Closes #16.

---

### Additional information

* The fix is simple but might not go necessarily into upcoming release.
* Also the #47 is a separate issue so this one looks fixed.


jodator:
> What about this PR?

I'll update it and make it work again.

jodator:
@Reinmar I've fixed it and make to code ready to be viewed by others ;)

PR #62, opened 2018-06-18
Fix: Selection should be in the first cell of an inserted table.

### LINKED ISSUE #61: Selection should be in the first cell of an inserted table
Right now it stays outside:

jun-18-2018 12-19-40

jodator:

Fix: Selection should be in the first cell of an inserted table. Closes #61.

---

### Additional information

* Simple fix - I've made `TableUtils#createTable()` return inserted table element so `InsertTableCommand` can set the selection. I don't think that `TableUtils#createTable()` method should set the selection.

### DIFF
--- a/src/commands/inserttablecommand.js
+++ b/src/commands/inserttablecommand.js
@@ -53,7 +53,11 @@ export default class InsertTableCommand extends Command {
 		const isRoot = firstPosition.parent === firstPosition.root;
 		const insertPosition = isRoot ? Position.createAt( firstPosition ) : Position.createAfter( firstPosition.parent );
 
-		tableUtils.createTable( insertPosition, rows, columns );
+		model.change( writer => {
+			const table = tableUtils.createTable( insertPosition, rows, columns );
+
+			writer.setSelection( Position.createAt( table.getChild( 0 ).getChild( 0 ) ) );
+		} );
 	}
 }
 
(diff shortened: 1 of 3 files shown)


src/tableutils.js
```diff
@@ -74,17 +74,22 @@ export default class TableUtils extends Plugin {
...
 	 * @param {Number} rows Number of rows to create.
 	 * @param {Number} columns Number of columns to create.
+	 * @returns {module:engine/model/element~Element} Created table element.
 	 */
 	createTable( position, rows, columns ) {
 		const model = this.editor.model;
 
+		let table;
+
 		model.change( writer => {
```
Reinmar: You can return table directly from `model.change()`. It fwds its callback's return value.

src/commands/inserttablecommand.js
```diff
@@ -53,7 +53,11 @@ export default class InsertTableCommand extends Command {
 		const isRoot = firstPosition.parent === firstPosition.root;
 		const insertPosition = isRoot ? Position.createAt( firstPosition ) : Position.createAfter( firstPosition.parent );
 
-		tableUtils.createTable( insertPosition, rows, columns );
+		model.change( writer => {
```
Reinmar: I miss a test for an integration with undo – if you'd use `model.change()` to wrap `writer.setSelection()`, then  (I think) you'd create two undo steps. Or at least – the created undo step might've been incorrect.

Reinmar: Try to write an integration test that undo is clear. If this test won't verify that `model.change()` needs to wrap both calls, then make it a unit test and check that a single batch was created.

Reinmar:
Minor R-.

PR #89, opened 2018-07-19
T/13: Implement table post-fxer

### LINKED ISSUE #13: Implement a table layout post-fixer
https://user-images.githubusercontent.com/156149/40480945-38d31f36-5f50-11e8-8358-54cca4e91c1b.gif

As you can see, after pressing <kbd>Backspace</kbd>, the table's layout is broken. The same may happen even if we'll fix the selection (#12) in collaboration scenarios. It can also happen on data load or pasting.

We need a post-fixer which will ensure that table layout is correct.

jodator:

Enhancement: Implement table post-fixer. Closes #13.

---

### Additional information

* This should cover all cases where table could be broken - so pasting and collaboration.
* @Mgsy the importan thing here is that Undo will be "kinda broken" but the table should always be correct: no missing table cells.
* The undo will create something that wasn't there before - kinda bug / kinda we cannot do better ATM

ps.: To the code reviewers: I might need to fix some internals / add some more comments before review.


src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+		}
+
+		// Fix table on any table's attribute change - including attributes of table cells.
+		if ( isTableAttributeEntry( entry ) ) {
+			table = getParentTable( entry.range.start );
+		}
+
+		if ( table ) {
+			// Step 1: correct rowspans of table cells if necessary.
+			wasFixed = fixTableCellsRowspan( table, writer ) || wasFixed; // Should be true if any of present tables was fixed.
```
oleq: I guess ` || wasFixed` is not necessary in this line. It's always `false`, as defined in L117.

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+/**
+ * @module table/converters/table-post-fixer
+ */
+
+import Position from '@ckeditor/ckeditor5-engine/src/model/position';
+import { getParentTable, updateNumericAttribute } from './../commands/utils';
+import TableWalker from './../tablewalker';
+
+/**
+ * Injects table post-fixer to the model.
```
oleq: > a table post-fixer

> into the model

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * The role of the tables post-fixer is to ensure that the table's rows has correct structure
```
oleq: > table post-fixer

>  table rows have the correct structure

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * after a {@link module:engine/model/model~Model#change `change()`} block was executed.
+ *
+ * The correct structure means that:
+ *
+ * * All table rows have the same width.
+ * * None of a table cells extends vertically beyond its section (either header or body).
```
oleq: > their section

> either a header or a body

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * If the table structure is not correct, the post-fixer will automatically correct it in two steps:
+ *
+ * 1. It will clip table cells that extends beyond it section.
```
oleq: > that extend

> their section

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * 2. It will add empty table cells to those rows which are narrower then the widest table row.
+ *
+ * ## Clipping overlapping table cells
+ *
+ * Such situation may occur when pasting a table (or part of a table) to the editor from external sources.
+ *
+ * As an example see below table which has a table cell (FOO) with a `rowspan` attribute of `2`:
```
oleq: > For example, see the following table which has the cell (FOO) with the `rowspan` attribute (`2`):

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *			</thead>
+ *			<tbody>
+ *				<tr>
+ *					<td>BAZ<td>
+ *					<td>XYZ<td>
+ *				</tr>
+ *			</tbody>
+ *		</table>
+ *
+ * In above example the table will be rendered as a table with two rows - one in the header and second one in the body.
```
oleq: > In the above example

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * The table cell with FOO contents will not expand to the body section so its `rowspan` attribute will be changed to `1` (and as
```
oleq: I don't get what this sentence is supposed to explain. Maybe some rephrase? There's also some parenthesis which is finished but without any start.

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * the value `1` is a default value the `rowspan` attribute will be removed from the model).
+ *
+ * The table cell with BAZ contents will be in the first column of a table.
```
oleq: > of the table

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * ## Adding missing table cells
+ *
+ * The table post-fixer will insert empty table cells to equalize table rows lengths. The width of a table row is calculated by counting
+ * widths of table cells and widths of columns spanned by cells from rows above in a given row.
```
oleq: > width of a table row

I'm not sure "width" is the best term here. It comes with a strong flavor of pixels or whatever measurement units.  In fact, you used "length" in the previous sentence :P

> from preceding rows

?

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * In the above example the table row in body section of the table is narrower then the row from the header - it has two cells
```
oleq: > example,

> in the body section

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * with default colspan (value of `1`). The header row has one cell with width = 1 and second withe width = 2.
```
oleq: > with the default colspan

> with the width

> s/withe/width

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * The table cell FOO does not expand beyond head section (and as such will be fixed in the first step of this post-fixer).
```
oleq: > the head section

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * The post-fixer will add a missing table cell to the row in body section of the table.
```
oleq: > in the body

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ *			</thead>
+ *			<tbody>
+ *				<tr>
+ *					<td>BAZ<td>
+ *					<td>XYZ<td>
+ *				</tr>
+ *			</tbody>
+ *		</table>
+ *
+ * **Note** The table post-fixer only ensures proper structure without deeper analysis of the nature of a change. As such it might lead
```
oleq: > As such,

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+ * to a structure which was not intended by user changes. In particular it will also fix undo steps (in conjunction with collaboration)
```
oleq: > by the user

> In particular,

TBH, I'd write a little bit more about the collaboration case. I think it's vital for the documentation.

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+			wasFixed = fixTableCellsRowspan( table, writer ) || wasFixed; // Should be true if any of present tables was fixed.
+			// Step 2: fix table rows widths.
+			wasFixed = fixTableRowsWidths( table, writer ) || wasFixed;
+		}
+	}
+
+	return wasFixed;
+}
+
+// Fixes the invalid value of rowspan attribute as a table cell cannot extend vertically beyond a table section to which it belongs.
```
oleq: > Fixes the invalid value of the rowspan attribute because a table cell cannot vertically extend beyond the table section it belongs to.

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+
+				wasFixed = true;
+			}
+		}
+	}
+
+	return wasFixed;
+}
+
+// Searches for the table cells that extends beyond the table section to which they belongs. It will return  an array of objects
```
oleq: > they belong to

Double whitespace before "an".

src/converters/table-post-fixer.js
```diff
@@ -0,0 +1,258 @@
...
+			lengths[ row ] = 0;
+		}
+
+		lengths[ row ] += 1;
+	}
+
+	return lengths;
+}
+
+// Checks if differ entry for attribute change is one of table's attributes.
```
oleq: > if the differ

> for an attribute

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+				root = model.document.getRoot();
+			} );
+	} );
+
+	afterEach( () => {
+		editor.destroy();
+	} );
+
+	describe( 'on insert table', () => {
+		it( 'should add missing columns to a tableRows that are shorter then longest table row', () => {
```
oleq: > to tableRows

> than the longest

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+			} );
+
+			expect( formatTable( getModelData( model, { withoutSelection: true } ) ) ).to.equal( formattedModelTable( [
+				[ '00', '', '' ],
+				[ '10', '11', '12' ],
+				[ '20', '21', '' ]
+			] ) );
+		} );
+
+		it( 'should add missing columns to a tableRows that are shorter then longest table row (complex 1)', () => {
```
oleq: ☝️

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+			} );
+
+			expect( formatTable( getModelData( model, { withoutSelection: true } ) ) ).to.equal( formattedModelTable( [
+				[ '00', { rowspan: 2, contents: '10' }, '', '' ],
+				[ '10', { colspan: 2, contents: '12' } ],
+				[ '20', '21', '', '' ]
+			] ) );
+		} );
+
+		it( 'should add missing columns to a tableRows that are shorter then longest table row (complex 2)', () => {
```
oleq: ☝️

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+			} );
+
+			expect( formatTable( getModelData( model, { withoutSelection: true } ) ) ).to.equal( formattedModelTable( [
+				[ { colspan: 6, contents: '00' } ],
+				[ { rowspan: 2, contents: '10' }, '11', { colspan: 3, contents: '12' }, '' ],
+				[ '21', '22', '', '', '' ]
+			] ) );
+		} );
+
+		it( 'should fix wrong rowspan attribute on table header', () => {
```
oleq: > the wrong

> of the table header

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+			} );
+
+			expect( formatTable( getModelData( model, { withoutSelection: true } ) ) ).to.equal( formattedModelTable( [
+				[ { rowspan: 2, contents: '00' }, { rowspan: 2, contents: '01' }, '02' ],
+				[ '12' ],
+				[ '20', '21', '22' ]
+			], { headingRows: 2 } ) );
+		} );
+
+		it( 'should fix wrong rowspan attribute on table body', () => {
```
oleq: ☝️

tests/converters/table-post-fixer.js
```diff
@@ -0,0 +1,426 @@
...
+				model.change( writer => {
+					writer.remove( Range.createIn( root ) );
+				} );
+			} ).to.not.throw();
+
+			expect( getModelData( model, { withoutSelection: true } ) ).to.equal( '<paragraph></paragraph>' );
+		} );
+	} );
+
+	describe( 'collaboration', () => {
```
oleq: The `it()` descriptions are not very helpful in this section. They should explain what the code is supposed to do, not simple ids of tests.

src/converters/table-post-fixer.js
jodator: You guessed wrong :) This case covers insertion of more then one table in one batch or inserting multiple table cells or rows in a batch.

In case that first table was fixed but the second one was not the `wasFixed` flag must be always `true` as there were fixes in whole batch.

ps.: I've updated the code a bit so a table will be analyzed only once in one batch.

PR #90, opened 2018-07-23
Fix: Switching heading columns/rows off will now behave as expected.

### LINKED ISSUE #34: Inconsequent behaviour when switching "table row" off
jun-08-2018 11-20-45

1. Make a table with 6 header rows.
2. Place the caret in the 3rd row.
3. Open the "row" options. The "header row" option is on. This is OK.
4. You click it to switch the header rows off.

Expected: Header rows should stay only on 1st and 2nd rows. 3-6 rows should be turned into normal rows.

Actual: Only rows 4-6 are changed. This doesn't make sense because it means that the "header row" option will still remain on for the 3rd row.

jodator:

Fix: Switching heading columns/rows off will now behave as expected. Closes #34.

---

### Additional information

* works as expected now :smile:


oleq:
image

Seems legit 😅

PR #97, opened 2018-08-01
T/56: Block content in table cells.

### LINKED ISSUE #56: Support block content inside tables
The [MVP](https://github.com/ckeditor/ckeditor5-table/issues/3) of the table feature doesn't contain a support for block content inside table cells. So, the first version will only support inline content (no images, no lists, no hard line breaks etc.)

Support for block content is our next priority after releasing the MVP. The challenge here is the UX. By default, we insert a table with inline content only:

```html
<table>
    <tr> 
       <td>Foo</td>
       <td>Boo</td>
   </tr>
</table>
```

When the user presses <kbd>Enter</kbd> in one of the cells, what should happen? We need to insert paragraphs there:

```html
<table>
    <tr> 
       <td><p>Foo</p><p>Bar</p></td>
       <td>Boo</td>
   </tr>
</table>
```

Is it ok that only one cell has block content and the other has inline content? This may cause vertical misalignment due to styling (paragraphs have margins) and general mess. So perhaps once you create block content in one of the cells, we should turn the entire table into "block table"?

There are also implementation issues. How to reflect that in the schema? A cell may contain inline content. Or it may contain block content. These are two different types of cells. Two different elements. Also, for features like headings, lists, etc. to be enabled, a block must be allowed in the current selection position. So, without paragraphs in the model all these features will be disabled. So, in fact, we always need paragraphs in the model.

And so on and so on :)

jodator:

Feature: Support block content inside table. Closes #56.

BREAKING CHANGE: Removed `table/commands/utils~getParentTable()` method. Use `table/commands/utils~findAncestor()` instead.

---

### Additional information

* This PR requires also change in the engine: https://github.com/ckeditor/ckeditor5-engine/pull/1489 - blocks inside `isLimit` elements should be selectable.
* Follow ups:
    * #96
    * #98 
    * #99


jodator:
@Mgsy have fun & try to break as much as possible with block contents in tables :)

src/tableutils.js
```diff
@@ -569,7 +569,9 @@ function createEmptyRows( writer, table, insertAt, rows, tableCellToInsert, attr
 // @param {module:engine/model/position~Position} insertPosition
 function createCells( cells, writer, insertPosition, attributes = {} ) {
 	for ( let i = 0; i < cells; i++ ) {
-		writer.insertElement( 'tableCell', attributes, insertPosition );
+		const tableCell = writer.createElement( 'tableCell', attributes );
+		writer.insert( tableCell, insertPosition );
```
oskarwrobel: You can insert `tableCell` to the `document` at the end (after inserting `paragraph` to the `tableCell`). This will create one less document operation.

oskarwrobel:
`Align Left` does not work in the table cell. Align feature assumes that default alignment is left and but for table call is center.

src/commands/setheaderrowcommand.js
```diff
@@ -171,7 +170,9 @@ function splitHorizontally( tableCell, headingRows, writer ) {
 		if ( columnIndex !== undefined && columnIndex === column && row === endRow ) {
 			const tableRow = table.getChild( row );
 
-			writer.insertElement( 'tableCell', attributes, Position.createFromParentAndOffset( tableRow, cellIndex ) );
+			const newCell = writer.createElement( 'tableCell', attributes );
+			writer.insert( newCell, Position.createFromParentAndOffset( tableRow, cellIndex ) );
+			writer.insertElement( 'paragraph', Position.createAt( newCell, 0 ) );
```
oskarwrobel: The same as https://github.com/ckeditor/ckeditor5-table/pull/97/files#r206859198

src/commands/utils.js
```diff
@@ -8,16 +8,27 @@
  */
 
 /**
- * Returns the parent table.
+ * Returns the parent table. Returns undefined if position is not inside table.
  *
  * @param {module:engine/model/position~Position} position
  * @returns {module:engine/model/element~Element|module:engine/model/documentfragment~DocumentFragment}
  */
 export function getParentTable( position ) {
```
oskarwrobel: Is this alias really helpful?

src/converters/table-post-fixer.js
```diff
@@ -304,7 +304,9 @@ function fixTableRowsSizes( table, writer ) {
 
 			if ( columnsToInsert ) {
 				for ( let i = 0; i < columnsToInsert; i++ ) {
-					writer.insertElement( 'tableCell', Position.createAt( table.getChild( rowIndex ), 'end' ) );
+					const tableCell = writer.createElement( 'tableCell' );
+					writer.insert( tableCell, Position.createAt( table.getChild( rowIndex ), 'end' ) );
+					writer.insertElement( 'paragraph', tableCell );
```
oskarwrobel: https://github.com/ckeditor/ckeditor5-table/pull/97/files#r206859198

src/converters/tablecell-post-fixer.js
```diff
@@ -0,0 +1,138 @@
...
+				const singleChild = tableCell.getChild( 0 );
+				const renameTo = Array.from( singleChild.getAttributes() ).length ? 'p' : 'span';
+
+				renameParagraphIfDifferent( singleChild, renameTo, writer, mapper );
+			}
+		} else {
+			// Check all nodes inside table cell on insert/remove operations (also other blocks).
+			const tableCell = entry.position && entry.position.parent;
+
+			if ( tableCell && tableCell.is( 'tableCell' ) ) {
```
oskarwrobel: `tableCell.is( 'tableCell' )` 😉

src/converters/upcasttable.js
```diff
@@ -53,9 +59,11 @@ export default function upcastTable() {
 			} else {
 				// Create one row and one table cell for empty table.
 				const row = conversionApi.writer.createElement( 'tableRow' );
-
 				conversionApi.writer.insert( row, ModelPosition.createAt( table, 'end' ) );
-				conversionApi.writer.insertElement( 'tableCell', ModelPosition.createAt( row, 'end' ) );
+
+				const tableCell = conversionApi.writer.createElement( 'tableCell' );
+				conversionApi.writer.insert( tableCell, ModelPosition.createAt( row, 'end' ) );
+				conversionApi.writer.insertElement( 'paragraph', ModelPosition.createAt( tableCell, 'end' ) );
```
oskarwrobel: https://github.com/ckeditor/ckeditor5-table/pull/97/files#r206859198

src/converters/tablecell-post-fixer.js
jodator: I could name it `possibleTableCell` ;P

oskarwrobel: Or `parent`. But `tableCell` is fine too.

src/converters/upcasttable.js
```diff
@@ -85,6 +93,50 @@ export default function upcastTable() {
...
+				// Should end after but we need to take into consideration that children could split our
+				// element, so we need to move range after parent of the last converted child.
+				// before: <allowed>[]</allowed>
+				// after: <allowed>[<converted><child></child></converted><child></child><converted>]</converted></allowed>
+				ModelPosition.createAfter( tableCell )
+			);
+
+			// Continue after inserted element.
+			data.modelCursor = data.modelRange.end;
+		}, { priority: 'normal' } );
```
oskarwrobel: Normal is a default value, so I would not define it. It's a detail but it's a few bytes of code more.

oskarwrobel:
The code looks good to me. I found a few places where you can reduce the number of created operations (it should matter for insertRow and insertColumn commands where many operations are created at once).

jodator:
> Align Left does not work in the table cell. Align feature assumes that default alignment is left and but for table call is center.

Yeah - I've already pinged [oleq about this](https://github.com/ckeditor/ckeditor5-table/issues/56#issuecomment-410171517).

src/commands/utils.js
jodator: Not now - it was before, but I'd remove it as the `getParentElement( 'tableCell' )` is more helpfull. I'll ook also if any existing engine method is not better here...

Mgsy:
## Steps to reproduce
1. Go to the `table block content` sample.
1. Put the caret in the last cell of the first column. 
1. Press <kbd>Tab</kbd> to select the image.
1. Press <kbd>Delete</kbd>.

## Current result
The editor crashes.

## Error
```
domconverter.js:280 Uncaught TypeError: Cannot read property 'parent' of null
    at DomConverter.viewRangeToDom (domconverter.js:280)
    at View.scrollToTheSelection (view.js:269)
    at Document.listenTo (delete.js:42)
    at Document.fire (emittermixin.js:196)
    at Document.DeleteObserver.document.on (deleteobserver.js:54)
    at Document.fire (emittermixin.js:196)
    at KeyObserver.fire (domeventobserver.js:96)
    at KeyObserver.onDomEvent (keyobserver.js:28)
    at ProxyEmitter.listenTo (domeventobserver.js:79)
    at ProxyEmitter.fire (emittermixin.js:196)
```

Mgsy:
## Steps to reproduce
1. Go to the table block content sample.
1. Select whole block element in some cell.
1. Press <kbd>Ctrl</kbd> + <kbd>X</kbd>.

## Current result
The table toolbar has disappeared. 

Browser: Edge, Chrome, Safari

In Firefox, the toolbar disappears for a second and shows up again.

Mgsy:
## Steps to reproduce
1. Go to the `table block content` sample.
1. Put the caret in the last cell of the first column.
1. Press <kbd>Tab</kbd> to select the image.
1. Copy it.
1. Paste it to same cell with a content.
1. Put the selection outside of this cell.
1. Navigate to this cell with <kbd>Tab</kbd>.

## Current result
Whole table's content is selected.

Browser: All browsers

Mgsy:
## Steps to reproduce
1. Go to the `table block content` sample.
1. Put the caret in the last cell of the first column.
1. Press <kbd>Tab</kbd> to select the image.
1. Repeat the previous step.

## Current result
The selection moves to the beginning of the editor.

jodator:
After some testing and talks - I'd remove Image support in tables for this MVP as widget-ception causes some problems. Added a follow-up: #99.

jodator:
> Steps to reproduce
> 
>     Go to the table block content sample.
>     Select whole block element in some cell.
>     Press Ctrl + X.
> 
> Current result
> 
> The table toolbar has disappeared.
> 
> Browser: Edge, Chrome, Safari
> 
> In Firefox, the toolbar disappears for a second and shows up again.

Looks like fixed...

jodator:
@oskarwrobel  & @Mgsy I've updated the PR with your comments i mind :)

PR #115, opened 2018-09-19
Align image toolbar to the widget toolbar repository

### LINKED ISSUE #107: Table toolbar improvements
https://github.com/ckeditor/ckeditor5-table/blob/34c8ee1336ed6c33e1a2c78c73d57d14c034f3be/src/tabletoolbar.js#L19 Shouldn't it be `when a table cell is select`? Because when the widget is selected the toolbar is hidden.

https://github.com/ckeditor/ckeditor5-table/blob/34c8ee1336ed6c33e1a2c78c73d57d14c034f3be/src/tabletoolbar.js#L50-L52 This comment is actually incorrect as table toolbar is visible only when the selection is inside the table. I don't know if the toolbar should be hidden then. Now you can still use shortcuts when the whole table is selected to e.g. bold, italic and link every cell in the table.

ma2ciek:

Other: Aligned `ImageToolbar` to the widget toolbar repository. Closes #107.

---

### Additional information

*For example – encountered issues, assumptions you had to make, other affected tickets, etc.*


src/tabletoolbar.js
```diff
@@ -41,123 +37,17 @@ export default class TableToolbar extends Plugin {
...
-		this._toolbar = new ToolbarView();
-
-		// Add buttons to the toolbar.
-		this._toolbar.fillFromConfig( toolbarConfig, editor.ui.componentFactory );
-
-		// Show balloon panel each time table widget is selected.
-		this.listenTo( editor.ui, 'update', () => {
-			this._checkIsVisible();
+		widgetToolbarRepository.register( 'table', {
+			toolbarItems: editor.config.get( 'table.toolbar' ) || [],
```
pjasiun: It should be called `table.tableToolbar`.

ma2ciek: I've split this PR into two parts, so we can align to the new API first and then add the new toolbar, add depreciation for `table.toolbar` and so on.

ma2ciek: But we can do all in one PR if you want

ma2ciek:
I wasn't sure in this and other PRs about tests, but I left them as they were because they check the whole integration pretty well.

PR #127, opened 2018-09-25
Fix: The table cell should always have paragraph in the model.

### LINKED ISSUE #125: Weird things start to happen after pasting a (disallowed) image into a table cell
sep-25-2018 14-42-18

You can't actually see this on the screencast due to low frame rate but after pasting an image to a table cell (image gets filtered out) you cannot any more place the caret in that table cell. When you click, the selection moves to another table cell.

jodator:

Fix: The table cell should always have paragraph in the model. Closes #125.

BREAKING CHANGE: The `injectTablePostFixer()` function from  `table/converters/table-post-fixer` is now `injectTableLayoutPostFixer()`and is moved to `table/converters/table-layout-post-fixer` module.

---

### Additional information

* I've added you @Reinmar for the review - I'm not sure if the fix is sufficient.

### DIFF
--- a/src/converters/table-cell-content-post-fixer.js
+++ b/src/converters/table-cell-content-post-fixer.js
@@ -1,0 +1,112 @@
+/**
+ * @license Copyright (c) 2003-2018, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md.
+ */
+
+/**
+ * @module table/converters/table-cell-content-post-fixer
+ */
+
+/**
+ * Injects a table cell post-fixer into the model.
+ *
+ * The role of the table post-fixer is to ensure that the table cells have the correct content
+ * after a {@link module:engine/model/model~Model#change `change()`} block was executed.
+ *
+ * A table cells must contains at least one block as a child. The empty table cell will have empty `<paragraph>` as a child.
+ *
+ *        <table>
+ *            <tableRow>
+ *                <tableCell></tableCell>
+ *            </tableRow>
+ *        </table>
+ *
+ * Will be fixed to:
+ *
+ *        <table>
+ *            <tableRow>
+ *                <tableCell><paragraph></paragraph></tableCell>
+ *            </tableRow>
+ *        </table>
+ *
+ * @param {module:engine/model/model~Model} model
+ */
+export default function injectTableCellContentPostFixer( model ) {
+	model.document.registerPostFixer( writer => tableCellContentsPostFixer( writer, model ) );
+}
+
+// The table cell contents post-fixer.
+//
+// @param {module:engine/model/writer~Writer} writer
+// @param {module:engine/model/model~Model} model
+function tableCellContentsPostFixer( writer, model ) {
+	const changes = model.document.differ.getChanges();
+
+	let wasFixed = false;
+
+	for ( const entry of changes ) {
+		// Enforce paragraph in tableCell even after other feature remove its contents.
+		if ( entry.type == 'remove' && entry.position.parent.is( 'tableCell' ) ) {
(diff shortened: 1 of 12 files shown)


src/converters/table-post-fixer.js
```diff
@@ -195,23 +196,41 @@ import TableWalker from './../tablewalker';
...
- *					<td>(empty, inserted by B)<td>
- *					<td>(empty, inserted by a post-fixer)<td>
+ *					<td><p>(empty, inserted by B)</p></td>
+ *					<td><p>(empty, inserted by B)</p></td>
+ *					<td><p>(empty, inserted by a post-fixer)</p></td>
  *				</tr>
  *			</tbody>
  *		</table>
  *
+ * ## Ensuring proper table structure
```
Reinmar: table structure? or rather "Ensuring a proper content of table cells"?

src/converters/table-post-fixer.js
```diff
@@ -49,14 +50,14 @@ import TableWalker from './../tablewalker';
  *		<table>
  *			<thead>
  *				<tr>
- *					<td rowspan="2">FOO</td>
- *					<td colspan="2">BAR</td>
+ *					<td rowspan="2"><p>FOO</p></td>
```
Reinmar: It seems that my comment disappeared so one more time...

Is it correct? Those `<p>`s won't be rendered (there will be spans or nothing). It's a bit confusing and I'm not sure if this is a mistake or a mental shortcut.

tests/converters/table-post-fixer.js
```diff
@@ -181,6 +181,19 @@ describe( 'Table post-fixer', () => {
 		} );
 	} );
 
+	describe( 'on removing tableCell contents', () => {
+		it( 'should fix empty table cells', () => {
```
Reinmar: It's hard to understand from this test that it checks that a paragraph should be added. I checked that it fails without the patch and the error is quite obvious then, but the code and description of this test are cryptic. 

There are many opinions regarding writing so many tools for creating tests. But, leaving this aside, when we write them we should make sure that tests remain understandable.

Reinmar:
Other than the issues in the code, I checked that this PR fixes the bug.

src/converters/table-post-fixer.js
jodator: I'd remove them.

jodator:
@Reinmar I've updated the solution:
- the Table Post-Ffixer was renamed to a more descriptive: Table Layout Post-Fixer as it fixes table layout (missing table cells)
- I've moved the table cell content post fixer to own file as it does something else then table layout post-fixer
- more tests added as it should act also on adding table/tableRow/tableCell to the model.

src/converters/table-cell-content-post-fixer.js
```diff
@@ -0,0 +1,112 @@
...
+
+	return wasFixed;
+}
+
+// Fixes all table cell content by adding paragraph to a table cell without any child.
+//
+// @param {module:engine/model/element~Element} table
+// @param {module:engine/model/writer~Writer} writer
+function fixTableCellContent( tableCell, writer ) {
+	if ( tableCell.childCount == 0 ) {
```
Reinmar: What if a cell already has some content, but that's a text? We can make this a followup because I don't know right now scenario which would lead to this, but it's certainly possible.

jodator: AFAIR the `$text` is not allowed in the model of a table cell.

Reinmar: Which does not mean it cannot end there. Schema rules are not enforced. Check out https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/deep-dive/schema.html#implementing-additional-constraints and the following section.

PR #129, opened 2018-09-27
Added an integration test for ckeditor/ckeditor5#1265

pomek:

Tests: Added an integration test for ckeditor/ckeditor5#1265.

Requires: https://github.com/ckeditor/ckeditor5-engine/pull/1559.

### DIFF
--- a/tests/manual/table.js
+++ b/tests/manual/table.js
@@ -9,10 +9,11 @@ import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor'
 import ArticlePluginSet from '@ckeditor/ckeditor5-core/tests/_utils/articlepluginset';
 import Table from '../../src/table';
 import TableToolbar from '../../src/tabletoolbar';
+import BlockQuote from '@ckeditor/ckeditor5-block-quote/src/blockquote';
 
 ClassicEditor
 	.create( document.querySelector( '#editor' ), {
-		plugins: [ ArticlePluginSet, Table, TableToolbar ],
+		plugins: [ ArticlePluginSet, Table, TableToolbar, BlockQuote ],
 		toolbar: [
 			'heading', '|', 'insertTable', '|', 'bold', 'italic', 'bulletedList', 'numberedList', 'blockQuote', 'undo', 'redo'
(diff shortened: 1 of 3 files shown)


tests/table-integration.js
```diff
@@ -78,6 +79,18 @@ describe( 'Table feature – integration', () => {
 				[ '<blockQuote><paragraph>bar[]</paragraph></blockQuote>' ]
 			] ) );
 		} );
+
+		it.only( 'merges elements without throwing errors', () => {
```
Reinmar: `.only()`

ma2ciek: We should have a test whether some test conains `/(describe|it)\.only\(/` on the CI 😄

pomek: Coverage drops below the "critical" level. It should be a warning for us. Unfortunately, @Reinmar was faster then I read the email :D

tests/manual/table.js
```diff
@@ -9,10 +9,11 @@ import ClassicEditor from '@ckeditor/ckeditor5-editor-classic/src/classiceditor'
 import ArticlePluginSet from '@ckeditor/ckeditor5-core/tests/_utils/articlepluginset';
 import Table from '../../src/table';
 import TableToolbar from '../../src/tabletoolbar';
+import BlockQuote from '@ckeditor/ckeditor5-block-quote/src/blockquote';
 
 ClassicEditor
 	.create( document.querySelector( '#editor' ), {
-		plugins: [ ArticlePluginSet, Table, TableToolbar ],
+		plugins: [ ArticlePluginSet, Table, TableToolbar, BlockQuote ],
```
Reinmar: Why was this change made? If you want to have a playground to test table with block quotes we'd also need this button in the toolbar.

pomek: image

The button is already added. https://github.com/ckeditor/ckeditor5-table/pull/129/files#diff-4097acfafae638fcb639ff3a9d047f3cR18

pomek: Yea, I was also a little surprised. I wanted to revert my invalid change than I saw the button :D

PR #151, opened 2018-11-27
T/134: Wrap $text nodes in paragraphs - table cell content post fixer addition

### LINKED ISSUE #134: Check table post fixers/converters with text in table cell
What if a cell already has some content, but that's a text?" problem
Follow up of:
https://github.com/ckeditor/ckeditor5-table/pull/127#pullrequestreview-160669503

Check if converters work properly with `#text` inside table cells.

jodator:

Fix: Add check for $text node in table cell content in post-fixer. Closes #134.

---

### Additional information

* As [pointed out by](https://github.com/ckeditor/ckeditor5-table/pull/127#discussion_r221924586) @Reinmar  the table cell can contain a `$text` node - ie created by other features as this is not enforced by schema.

### DIFF
--- a/src/converters/table-cell-content-post-fixer.js
+++ b/src/converters/table-cell-content-post-fixer.js
@@ -97,16 +97,28 @@ function fixTableRow( tableRow, writer ) {
 	return wasFixed;
 }
 
-// Fixes all table cell content by adding paragraph to a table cell without any child.
+// Fixes all table cell content by:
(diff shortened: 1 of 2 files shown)


Reinmar:
I think I'll merge this PR but it actually will need to be reverted soon as this is a more generic problem and we need a more generic solution: https://github.com/ckeditor/ckeditor5/issues/1464.

PR #176, opened 2019-02-21
Make table cell view post-fixer don't fix a valid selection.

jodator:

Fix: Table cell view post-fixer should not fix valid view selection. Closes ckeditor/ckeditor5#1554.

---

### Additional information

* Dunno if `viewSelection.getRangesCount` check is needed but if there are no renges in selection the `anchor` and `focus` are `null` so this check.


Reinmar:
Hm... I don't think that this is a complete fix. The model and view selections are correct after pasting this URL:

image

The model selection is on the media element and the view selection is on the widget. The widget is marked as selected, still we don't render a fake selection (there's no fake sel container in the DOM). So there's a bug somewhere else too.

I don't mean that there's something wrong in your PR – I didn't actually check it. I noticed that you focused on ckeditor5-table, while from the evidence of this bug, it was quite clear that there's a bug somewhere else (too).

jodator:
>  it was quite clear that there's a bug somewhere else (too).

I think that it might be other bug. But the problem with selection was because of table cell view post fixer which updated the selection always if there was renaming `<span>` to `<p>` or vice versa but it only should update the selection only if the view selection points to old renamed element.

ps.: this works on my machine :disappointed: - no double selection.

jodator:
ps.: The fake selection container is appended to editable root (in this case `td`) so again - for me this works: https://github.com/ckeditor/ckeditor5/issues/1554#issuecomment-466004438.

Reinmar:
> ps.: The fake selection container is appended to editable root (in this case `td`) so again - for me this works: [ckeditor/ckeditor5#1554 (comment)](https://github.com/ckeditor/ckeditor5/issues/1554#issuecomment-466004438).

I meant the situation on `master`. Your PR somehow makes the engine not enter this path. I tested `master` because I'm worried why this path doesn't work.

jodator:
@Reinmar & @mlewand I've updated the PR to the latest master. It looks like the issue with tables is fixed. I cannot reproduce it outside the tables with current configuration.

I've tried to check if the nested editable is a culprit - so I've enabled `<softBreak>` and `<media>` in `<caption>` and it works fine on master (no blinking caret):
Peek 2019-05-21 17-15

SO this looks like the issue was mostly with table and as such is fixed in this PR.

PR #180, opened 2019-04-01
T/ckeditor5/1620: Fix downcast conversion in data pipeline when converting single paragraph with attributes in table cell.

jodator:

Fix: Single paragraphs with attributes inside tableCell will be properly converted in data pipeline.. Closes ckeditor/ckeditor5#1620.

---

### Additional information

* It works but probably some refactor of those converters would be nice.


jodator:
@Mgsy  could you check this case?

PR #186, opened 2019-05-06
Table rows will not be added on tab key press if the associated command is disabled

### LINKED ISSUE #185: Adding a row on tab key should use a command
At the moment, if <kbd>Tab</kbd> key is pressed in the last cell of the table a new table row is created and the caret is moved there. This is "hardcoded" - doesn't use `insertTableRowBelow` command. This is incorrect as the command (and all other commands) might be disabled if the table is not editable for some reason (for example, comments-only mode). It will also be better for integration with track changes.

scofalik:

Fix: Table rows will not be added on tab key press if the associated command is disabled. Closes #185.

### DIFF
--- a/src/tableediting.js
+++ b/src/tableediting.js
@@ -220,7 +220,7 @@ export default class TableEditing extends Plugin {
 				editor.execute( 'insertTableRowBelow' );
 
 				// Re-evaluate `isLastRow`. If `insertTableRowBelow` execution didn't add any row (because it was disabled or it got
-				// overwritten in some way) this will still be `false`. But if the row was added it will change to `true`.
+				// overwritten in some way) this will still be `true`. But if the row was added it will change to `false`.
 				isLastRow = currentRowIndex === table.childCount - 1;
 			}
 
--- a/src/tableediting.js
+++ b/src/tableediting.js
@@ -214,25 +214,22 @@ export default class TableEditing extends Plugin {
 			}
 
 			const isLastCellInRow = currentCellIndex === tableRow.childCount - 1;
-			let isLastRow = currentRowIndex === table.childCount - 1;
+			const isLastRow = currentRowIndex === table.childCount - 1;
 
 			if ( isForward && isLastRow && isLastCellInRow ) {
 				editor.execute( 'insertTableRowBelow' );
 
-				// Re-evaluate `isLastRow`. If `insertTableRowBelow` execution didn't add any row (because it was disabled or it got
-				// overwritten in some way) this will still be `true`. But if the row was added it will change to `false`.
-				isLastRow = currentRowIndex === table.childCount - 1;
+				// Check if the command actually added a row. If `insertTableRowBelow` execution didn't add a row (because it was disabled
+				// or it got overwritten) do not change the selection.
+				if ( currentRowIndex === table.childCount - 1 ) {
+					return;
+				}
 			}
 
 			let cellToFocus;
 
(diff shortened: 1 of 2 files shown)


tests/tableediting.js
```diff
@@ -294,6 +294,22 @@ describe( 'TableEditing', () => {
 				] ) );
 			} );
 
+			it( 'should not create another row and not move the caret if insertTableRowBelow command is disabled', () => {
+				setModelData( model, modelTable( [
+					[ '11', '[12]' ]
```
jodator: maybe test should include also testing the selection so it should start with collapsed and should stay collapsed (not changed) after the <kbd>tab</kbd>.

jodator: maybe test should include also testing the selection so it should start with collapsed and should stay collapsed (not changed) after the <kbd>tab</kbd>.

src/tableediting.js
```diff
@@ -214,16 +214,25 @@ export default class TableEditing extends Plugin {
...
 
 			const isLastCellInRow = currentCellIndex === tableRow.childCount - 1;
-			const isLastRow = currentRowIndex === table.childCount - 1;
+			let isLastRow = currentRowIndex === table.childCount - 1;
 
 			if ( isForward && isLastRow && isLastCellInRow ) {
-				editor.plugins.get( 'TableUtils' ).insertRows( table, { at: table.childCount } );
+				editor.execute( 'insertTableRowBelow' );
+
+				// Re-evaluate `isLastRow`. If `insertTableRowBelow` execution didn't add any row (because it was disabled or it got
```
jodator: Too complex? Maybe check if the command is enabled - if not just return.

src/tableediting.js
```diff
@@ -214,16 +214,25 @@ export default class TableEditing extends Plugin {
...
+				// overwritten in some way) this will still be `true`. But if the row was added it will change to `false`.
+				isLastRow = currentRowIndex === table.childCount - 1;
 			}
 
 			let cellToFocus;
 
 			// Move to first cell in next row.
 			if ( isForward && isLastCellInRow ) {
+				if ( isLastRow ) {
```
jodator: I don't like the `if` in `if` here. I'd move that check as `if( !command.isEnabled ) { return }` to the previous check of inserting table row.

src/tableediting.js
jodator: I am not sure what is the case for "got it overwritten in some way" part so maybe I'm wrong here.

But still if the number of rows didn't change I'd exit here rather then in focus condition below.

scofalik: That was my first idea, but I wanted to check the content state which is the ultimate source of truth and the indicator of what to do next.

By "got overwritten" I mean that commands behavior (execution) may be overwritten by other features. The command might be enabled but for some reason, it might not add a row. Track changes overwrites commands a lot and some of them do not change the content after execution. I am not yet sure how integration with tables will look like (now you can only add a table but table utils are not supported). So I think I am on a safe side here with the way I coded it.

tests/tableediting.js
scofalik: I copied this from the test above. I can change it to a collapsed selection.

src/tableediting.js
jodator: So the part about earlier abort still applies - just return here not below. (number of rows remains the same). WDYT?

scofalik: But I agree that this is confusing.

PR #188, opened 2019-05-10
Remove deprecated configuration option.

### LINKED ISSUE #167: Remove support for deprecated `table.toolbar`.
In the next major release, we should remove that property from being supported.

msamsel:

Other: Remove deprecated 'table.toolbar' configuration option. Closes #167.

BREAKING CHANGE: `config.table.toolbar` is now removed from code. Use `config.table.contentToolbar`instead.

---

### Additional information

~I remain configuration docs entry for this option. I think there should be some trace that such option existed. However Umberto doesn't mark it properly.~ It was removed as well after review comment.
Related issue ( https://github.com/cksource/umberto/issues/799 )

### DIFF
--- a/tests/tabletoolbar.js
+++ b/tests/tabletoolbar.js
@@ -238,59 +238,6 @@ describe( 'TableToolbar', () => {
 		} );
 	} );
 
-	describe( 'deprecated toolbar', () => {
-		let editor, editorElement;
-
-		it( 'table.toolbar should not create toolbar any longer', () => {
-			editorElement = global.document.createElement( 'div' );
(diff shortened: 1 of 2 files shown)


tests/tabletoolbar.js
```diff
@@ -239,12 +239,11 @@ describe( 'TableToolbar', () => {
 	} );
 
 	describe( 'deprecated toolbar', () => {
```
mlewand: To be honest I don't think we need unit tests for removing a variable. Instead we should remove the suite all along.

msamsel:
During this task I also found missing docs. I made a separate PR for that #189.

PR #192, opened 2019-05-20
Table cell view post-fixer will not crash if an element inside a cell got attribute and was removed at the same time

### LINKED ISSUE #191: Table cell view post-fixer crashes when element was changed and removed
Related to https://github.com/ckeditor/ckeditor5-table/issues/187.

In short, when a view element which is inside a table gets or has an attribute removed (or the model element was renamed), the view post-fixer will want to perform some action on view element bound to it.

However, if the model element was also removed, the model and view elements will not be mapped anymore, which causes a crash in the post-fixer.

Post-fixer should not fire for view elements which are removed.

scofalik:

Fixed: Table cell view post-fixer will not crash if an element inside a cell got attribute and was removed at the same time. Closes #191.

### DIFF
--- a/src/converters/tablecell-post-fixer.js
+++ b/src/converters/tablecell-post-fixer.js
@@ -126,7 +126,8 @@ function getElementsToCheck( view ) {
 // - span: for single paragraph with no attributes.
 // - p   : in other cases.
 function ensureProperElementName( currentViewElement, mapper, writer ) {
-	// This situation may happen if a view element was changed and removed at the same time. In this case the view element is already unbound so the post-fixer would crash.
+	// This situation may happen if a view element was changed and removed at the same time.
+	// In this case, the view element is already unbound so the post-fixer would crash.
 	if ( !currentViewElement.root.is( 'rootElement' ) ) {
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


src/converters/tablecell-post-fixer.js
```diff
@@ -126,6 +126,10 @@ function getElementsToCheck( view ) {
 // - span: for single paragraph with no attributes.
 // - p   : in other cases.
 function ensureProperElementName( currentViewElement, mapper, writer ) {
+	if ( !currentViewElement.root.is( 'rootElement' ) ) {
```
pjasiun: Could you leave here a comment about how root can not be `rootElement` and why `false` should be returned in such case?

scofalik:
```suggestion
	// This situation may happen if a view element was changed and removed at the same time. In this case the view element is already unbound so the post-fixer would crash.
	if ( !currentViewElement.root.is( 'rootElement' ) ) {
```

scofalik: Done.

src/converters/tablecell-post-fixer.js
```diff
@@ -126,6 +126,11 @@ function getElementsToCheck( view ) {
 // - span: for single paragraph with no attributes.
 // - p   : in other cases.
 function ensureProperElementName( currentViewElement, mapper, writer ) {
+	// This situation may happen if a view element was changed and removed at the same time. In this case the view element is already unbound so the post-fixer would crash.
```
scofalik:
```suggestion
	// This situation may happen if a view element was changed and removed at the same time.
	// In this case, the view element is already unbound so the post-fixer would crash.
```

PR #195, opened 2019-07-05
Changed view table cell post-fixer to a model table cell post-fixer

### LINKED ISSUE #187: Table cell post fixer considered harmful
We use a post fixer set on view to manage the content of table cells. We have this rule that, depending on the table cell content, it should be paragraphed or not. But to make things easier in the model, we keep paragraphs in the model.

Basically, what the post-fixer does is this: for every changed table cell content, check all top-level elements in it. If it is just one `<paragraph>` with no attributes, rename its view element to `<span>` (if it is not a `<span>`). In other case, rename it to `<p>` if it is a `<span>`.

Renaming is removing + inserting a new element. This have an important impact on other features. It is dangerous to make such changes outside of conversion mechanism. A feature might have a reference to the view element and now it is gone which leads to errors. For example, there might have been a marker connected with that element.

It would be better if we had a solution that does not operate directly on the view.

scofalik:

Internal: Changed view table cell post-fixer to a model table cell post-fixer. Closes #187.


scofalik:
Related PR in engine: https://github.com/ckeditor/ckeditor5-engine/pull/1756

PR #196, opened 2019-07-11
T/ckeditor5/1151: Adjusted `InsertTableView` for better compatibility with right–to–left (RTL) languages

oleq:

Other: Adjusted `InsertTableView` for better compatibility with right–to–left (RTL) languages. See ckeditor/ckeditor5#1151.

---

### Additional information

A piece of https://github.com/ckeditor/ckeditor5/pull/1881.


Reinmar:
I don't understand this change. Can you explain why spaces were bad?

Reinmar:
There's also https://github.com/ckeditor/ckeditor5/issues/1151#issuecomment-515064624 which perhaps we can squeeze into this PR.

oleq:
When you have a table let's say 2x5 (LTR) it is 2 rows and 5 columns. But when you enable RTL UI things go backwards and the label is displayed 5x2, which makes no sense. Maybe I should've used nbsp instead. I made this change at the beginning and didn't think it through.

> There's also ckeditor/ckeditor5#1151 (comment) which perhaps we can squeeze into this PR (if not, let's make it a followup for it27).

Follow–up.

oleq:
Sorted out the label problem. Used the multiplication sign '×' instead of 'x' letter and it looks great now.

Reinmar:
Interesting problem :D I didn't think about this at all.

PR #202, opened 2019-08-13
Improvements for integration with track changes

scofalik:

Feature: `TableWalker` will now return `cell` value also for spanned cells when traversing a table with `includeSpanned` option set to `true`. Additionally, `isSpanned` property was introduced in returned values.
Internal: Improvements in table plugin allowing for integration with track changes.

BREAKING CHANGE: `TableWalker` will not return `undefined` as `cell` value for spanned cells anymore. Use `isSpanned` instead.

---

### Additional information

Requires https://github.com/ckeditor/ckeditor5-engine/pull/1781


src/tableutils.js
```diff
@@ -219,13 +219,16 @@ export default class TableUtils extends Plugin {
...
 
-			for ( const { row, column, cell, colspan, rowspan, cellIndex } of tableWalker ) {
+			for ( const { row, cell, cellIndex } of tableWalker ) {
 				// When iterating over column the table walker outputs either:
 				// - cells at given column index (cell "e" from method docs),
 				// - spanned columns (spanned cell from row between cells "g" and "h" - spanned by "e", only if `includeSpanned: true`),
 				// - or a cell from the same row which spans over this column (cell "a").
 
-				if ( column !== insertAt ) {
+				const rowspan = parseInt( cell.getAttribute( 'rowspan' ) || 1 );
```
jodator: why not take it from table walker if already available?

src/tablewalker.js
```diff
@@ -341,59 +325,69 @@ export default class TableWalker {
 		return rowSpans.has( column );
 	}
 
+	/**
+	 * Returns the cell element that is spanned at `row` x `column` location.
```
jodator: over?

tests/tablewalker.js
```diff
@@ -157,29 +157,29 @@ describe( 'TableWalker', () => {
...
-		it( 'should output spanned cells as empty cell', () => {
+		it( 'should output spanned cells', () => {
 			testWalker( [
 				[ { colspan: 2, rowspan: 3, contents: '00' }, '02' ],
 				[ '12' ],
 				[ '22' ],
 				[ '30', { colspan: 2, contents: '31' } ]
 			], [
 				{ row: 0, column: 0, index: 0, data: '00' },
-				{ row: 0, column: 1, index: 1, data: undefined },
```
jodator: I see what you did with the walker and I now the docs are not in line with the `includedSpaned` intention.

The intention was to iterate either over existing table cells or by virtual (empty) places. For such empty places we can't output table cell as it will be in conflict with that use case.

The `include` spanned option was helpful to manage table modifications (extending table cells on a column, row, etc).

If you see the `cell` to be needed in the value we must indicate that current value is `spanned` by other cell.

However I can see that this is not used that much (this empty cell detection) and all the other features works o_0.

Anyway I think that we should keep this information - that a currently yielded table walker value belongs to a place spanned by other cell (so maybe `isSpanned: true` will be OK?

scofalik:
As far as `includeSpanned` and `cell` value. (I think) I understand the intention behind `includeSpanned` and it worked fine but in my case, I needed to get to the cell that is spanning over given table location. For example, when I was iterating over all locations in a column, I wanted to get both cells that "starts" in that column (those were added to a suggestion for sure) but also I wanted to store spanned cells because they might need to be added to a suggestion too (upon further processing).

I've checked if changing `cell` from `undefined` breaks anything and it didn't (so no scripts depended on `undefined` value). So, for me, not passing a value there is wasting an information that we have for free. I am fine for adding `isSpanned` though, for theoretical future needs.

jodator:
> I am fine for adding `isSpanned` though, for theoretical future needs.

So let's do this - we shouldn't loose that information. Also update the commit message regarding this change.

src/commands/setheadercolumncommand.js
```diff
@@ -61,21 +61,27 @@ export default class SetHeaderColumnCommand extends Command {
 	 * When the selection is already in a header column, it will set `headingColumns` so the heading section will end before that column.
 	 *
 	 * @fires execute
+	 * @params {Boolean} [forceValue] If set, the command will set (`true`) or unset (`false`) header columns according to `forceValue`
+	 * parameter instead of the current model state.
 	 */
-	execute() {
+	execute( forceValue = null ) {
```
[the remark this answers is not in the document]
scofalik: We already used that name in block quote and attribute command so I just used it here. Basically, yes, if you use the command with `forceValue` set, the command will set that value. It boils down to whether anything should happen at all (in case of boolean command value) but it is needed to create suggestions that aren't dependent on the current content state but the state that was at the time of creating a suggestion.

I mean, I hoped that docs are clear enough :)

```
	 * @params {Boolean} [forceValue] If set, the command will set (`true`) or unset (`false`) header columns according to `forceValue`
	 * parameter instead of the current model state.
```

src/tableutils.js
scofalik: The story of this change is as follows:

1. Before any changes in table repository, it worked. (For some reason :P).
2. After I did some changes, there were bugs with inserting columns.
3. I debugged this part and found out that `colspan` and `rowspan` are always `1` when `includeSpanned: true`. That was causing issues. Actually, the docs says that `colspan` and `rowspan` should be `undefined` when using `includeSpanned`.
4. After I ended working on this PR and got a good understanding how the table walker works, I wanted to revert those changes and see if it is fine, but it wasn't.

So, to sum up:

1. I am not sure what exactly made this to fail, but...
2. It was using properties that were incorrectly set anyway.

scofalik: I also think, that maybe during a lot of "clicking" through different insertions with merged cells I might stumbled upon a case that was not in the original test suite and needed to re-work it for that case 🤷‍♂but this is speculation :).

src/commands/setheaderrowcommand.js
```diff
@@ -60,8 +60,10 @@ export default class SetHeaderRowCommand extends Command {
 	 * When the selection is already in a header row, it will set `headingRows` so the heading section will end before that row.
 	 *
 	 * @fires execute
+	 * @params {Boolean} [forceValue] If set, the command will set (`true`) or unset (`false`) header rows according to `forceValue`
+	 * parameter instead of the current model state.
 	 */
-	execute() {
+	execute( forceValue = null ) {
```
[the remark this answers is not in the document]
scofalik: It could be `execute( value )` or `execute( { value } )`. Note that `value` is essentially the same as `forceValue`. For me, `forceValue` makes more sense if usually the `value` is taken basing on the editor state. Also, as I wrote above, we already used that name.

tests/tablewalker.js
```diff
@@ -202,25 +203,25 @@ describe( 'TableWalker', () => {
...
-				{ row: 1, column: 0, index: 0, data: undefined },
-				{ row: 1, column: 1, index: 0, data: undefined },
-				{ row: 1, column: 2, index: 0, data: '12' },
-				{ row: 2, column: 0, index: 0, data: undefined },
-				{ row: 2, column: 1, index: 0, data: undefined },
-				{ row: 2, column: 2, index: 0, data: '22' }
+				{ row: 1, column: 0, index: 0, data: '00', isSpanned: true },
+				{ row: 1, column: 1, index: 0, data: '00', isSpanned: true },
+				{ row: 1, column: 2, index: 0, data: '12', isSpanned: false },
+				{ row: 2, column: 0, index: 0, data: '00', isSpanned: true },
```
jodator: WDYT about making `isSpanned: false` default value in walker test?

If set would stand-out more:
```js
{ row: 1, column: 0, index: 0, data: '00', isSpanned: true },
{ row: 1, column: 1, index: 0, data: '00', isSpanned: true },
{ row: 1, column: 2, index: 0, data: '12' },
{ row: 2, column: 0, index: 0, data: '00', isSpanned: true }
{ row: 2, column: 2, index: 0, data: '22' }
```

scofalik:
Okay, I updated docs and fixed what we agreed on fixing.

src/tablewalker.js
```diff
@@ -54,8 +55,8 @@ export default class TableWalker {
 	 *
 	 *		const tableWalker = new TableWalker( table, { startRow: 1, endRow: 1, includeSpanned: true } );
 	 *
-	 *		for ( const cellInfo of tableWalker ) {
-	 *			console.log( 'Cell at ' + cellInfo.row + ' x ' + cellInfo.column + ' : ' + ( cellInfo.cell ? 'has data' : 'is spanned' ) );
+	 *		for ( const value of tableWalker ) {
+	 *			console.log( 'Cell at ' + value.row + ' x ' + value.column + ' : ' + ( value.isSpanned ? 'has data' : 'is spanned' ) );
```
jodator:
```suggestion
	 *			console.log( 'Cell at ' + value.row + ' x ' + value.column + ' : ' + ( value.isSpanned ? 'is spanned' : 'has data' ) );
```

jodator:
Waits for: https://github.com/ckeditor/ckeditor5-engine/pull/1781.

PR #212, opened 2019-08-22
T/209:  Table cell post-fixer will refresh a cell only when it is needed.

### LINKED ISSUE #209: Cell post-fixer should refresh a cell only when it is needed
At the moment, the new cell post-fixer is pretty loose/naive. It checks only if something changed just inside a table cell and, if so, it refreshes the table cell. It would be better if it fixed only when really needed.

The way it is now is that it causes unnecessary cell refreshes when an image is uploaded, after each `uploadStatus` attribute change. This (together with another mistake) causes a serious bug.

To avoid unnecessary changes (and loops through `Differ`) and avoid bugs in future, the post-fixer should have a better condition applied.

jodator:

Fix: Table cell post-fixer will refresh a cell only when it is needed. Closes #209.

---

### Additional information

* Needed by: https://github.com/ckeditor/ckeditor5-image/pull/316 but also can be merged anyway. as a stand-alone PR.

ps.: I need to add some tests for this.

### DIFF
--- a/src/converters/table-cell-refresh-post-fixer.js
+++ b/src/converters/table-cell-refresh-post-fixer.js
@@ -67,12 +67,6 @@ function checkRefresh( tableCell, type ) {
 	}
 
 	const children = Array.from( tableCell.getChildren() );
-	const hasInnerText = children.some( child => child.is( 'text' ) );
-
-	// If a bare text node (not wrapped in a paragraph) was added - refresh it.
-	if ( hasInnerText ) {
-		return true;
-	}
 
 	const hasInnerParagraph = children.some( child => child.is( 'paragraph' ) );
 
--- a/src/converters/table-cell-refresh-post-fixer.js
+++ b/src/converters/table-cell-refresh-post-fixer.js
@@ -66,9 +66,7 @@ function checkRefresh( tableCell, type ) {
 		return true;
 	}
 
-	const children = Array.from( tableCell.getChildren() );
-
-	const hasInnerParagraph = children.some( child => child.is( 'paragraph' ) );
+	const hasInnerParagraph = Array.from( tableCell.getChildren() ).some( child => child.is( 'paragraph' ) );
 
 	// If there is no paragraph in table cell then the view doesn't require refreshing.
 	if ( !hasInnerParagraph ) {
--- a/src/converters/table-cell-refresh-post-fixer.js
+++ b/src/converters/table-cell-refresh-post-fixer.js
@@ -26,18 +26,18 @@ function tableCellRefreshPostFixer( model ) {
 	const differ = model.document.differ;
 
 	// Stores cells to be refreshed so the table cell will be refreshed once for multiple changes.
-	const cellToRefresh = new Set();
+	const cellsToRefresh = new Set();
 
 	for ( const change of differ.getChanges() ) {
 		const parent = change.type == 'insert' || change.type == 'remove' ? change.position.parent : change.range.start.parent;
 
 		if ( parent.is( 'tableCell' ) && checkRefresh( parent, change.type ) ) {
-			cellToRefresh.add( parent );
+			cellsToRefresh.add( parent );
 		}
 	}
 
-	if ( cellToRefresh.size ) {
-		for ( const tableCell of cellToRefresh.values() ) {
+	if ( cellsToRefresh.size ) {
+		for ( const tableCell of cellsToRefresh.values() ) {
 			differ.refreshItem( tableCell );
 		}
 
@@ -47,16 +47,16 @@ function tableCellRefreshPostFixer( model ) {
 	return false;
 }
 
-// Check if the model table cell requires refreshing to be re-rendered to a proper state in the view.
+// Checks if the model table cell requires refreshing to be re-rendered to a proper state in the view.
 //
-// This methods detects changes that will require renaming <span> to <p> (or vice versa) in the view,
+// This methods detects changes that will require renaming <span> to <p> (or vice versa) in the view.
 //
-// This method is simple heuristic that check only single change and will give false positive result when many changes will result
-// in a state that does not require renaming in the view.
+// This method is a simple heuristic that checks only a single change and will sometimes give a false positive result when multiple changes
+// will result in a state that does not require renaming in the view (but will be seen as requiring a refresh).
 //
-// For instance: The <span> to <p> should be renamed when adding an attribute to a <paragraph>.
-// But adding one attribute and removing another will result in false positive: the check for added attribute will see one attribute
-// on a paragraph and will falsy qualify such change as adding attribute to a paragraph without any attribute.
+// For instance: a `<span>` should be renamed to `<p>` when adding an attribute to a `<paragraph>`.
+// But adding one attribute and removing another one will result in a false positive: the check for added attribute will see one attribute
+// on a paragraph and will falsy qualify such change as adding an attribute to a paragraph without any attribute.
 //
 // @param {module:engine/model/element~Element} tableCell Table cell to check.
 // @param {String} type Type of change.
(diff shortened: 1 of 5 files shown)


src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
 
 	return fixed;
 }
+
+// Check if a table cell in the view requires refreshing.
```
scofalik:
```suggestion
// // Check if the model table cell requires refreshing to be re-rendered to a proper state in the view.
```

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
...
+// - <span> to <p> rename in the view,
+// - adding a missing <paragraph>,
+// - or wrapping a text node in <paragraph>
+//
+// thus requiring refreshing the table cell view.
+//
+// @param {module:engine/model/element~Element} tableCell Table cell to check.
+// @param {String} type Type of change.
+function checkRefresh( tableCell, type ) {
+	// If children of a table cell were removed - refresh it.
```
scofalik:
```suggestion
	// If all children of a table cell were removed - refresh it.
```

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
...
+	if ( !tableCell.childCount ) {
+		return true;
+	}
+
+	const children = Array.from( tableCell.getChildren() );
+	const hasInnerText = children.some( child => child.is( 'text' ) );
+
+	// If a bare text node (not wrapped in a paragraph) was added - refresh it.
+	if ( hasInnerText ) {
```
scofalik: Do we have a case when this situation occurs? We fix the table cell with paragraph ASAP, so I am not sure when this could happen 🤔

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
...
+
+	const hasInnerParagraph = children.some( child => child.is( 'paragraph' ) );
+
+	// If there is no paragraph in table cell then the view doesn't require refreshing.
+	if ( !hasInnerParagraph ) {
+		return false;
+	}
+
+	// For attribute change we only refresh single paragraphs as they might trigger <span> to <p> change in the view.
+	if ( type == 'attribute' ) {
```
scofalik: This will be refreshing for a paragraph that gets second and next attributes. Maybe not a big deal but still something.

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
...
+
+	// If a bare text node (not wrapped in a paragraph) was added - refresh it.
+	if ( hasInnerText ) {
+		return true;
+	}
+
+	const hasInnerParagraph = children.some( child => child.is( 'paragraph' ) );
+
+	// If there is no paragraph in table cell then the view doesn't require refreshing.
+	if ( !hasInnerParagraph ) {
```
scofalik: What if there was a clean paragraph (`<span>`) and it was replaced by an image?

tests/converters/table-cell-paragraph-post-fixer.js
```diff
@@ -221,4 +229,58 @@ describe( 'Table cell paragraph post-fixer', () => {
...
+			'<table>' +
+				'<tableRow>' +
+					'<tableCell><block></block></tableCell>' +
+				'</tableRow>' +
+			'</table>'
+		);
+
+		postFixerSpy.resetHistory();
+
+		// Insert table row with one table cell
```
scofalik: Copypaaaasta

tests/converters/table-cell-paragraph-post-fixer.js
```diff
@@ -221,4 +229,58 @@ describe( 'Table cell paragraph post-fixer', () => {
...
+
+		expect( formatTable( getModelData( model, { withoutSelection: true } ) ) ).to.equal( formatTable(
+			'<table>' +
+				'<tableRow>' +
+					'<tableCell><block foo="bar"></block></tableCell>' +
+				'</tableRow>' +
+			'</table>'
+		) );
+
+		sinon.assert.calledOnce( postFixerSpy );
```
scofalik: Should "not be run" but `.calledOnce()`?

scofalik: Ugh, okay, this is convoluted :|. Called once means that no other post-fixer was called. I wonder if we can spy this in a better way (`Differ#refreshItem` might be a little less bullet-proof though). :(

scofalik: OTOH, `calledTwice` is not bullet-proof either.

tests/converters/table-cell-paragraph-post-fixer.js
```diff
@@ -221,4 +229,58 @@ describe( 'Table cell paragraph post-fixer', () => {
...
+	} );
+
+	it( 'should be run on changing attribute of an paragraph in a table cell', () => {
```
scofalik:
```suggestion
	it( 'should be run on changing attribute of a paragraph in a table cell', () => {
```

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -39,3 +39,46 @@ function tableCellRefreshPostFixer( model ) {
...
+
+	// For attribute change we only refresh single paragraphs as they might trigger <span> to <p> change in the view.
+	if ( type == 'attribute' ) {
+		return tableCell.childCount === 1;
+	}
+
+	// For other changes (insert/remove) the <span> to <p> change can occur when:
+	// - sibling is added to a single paragraph (childCount == 2)
+	// - sibling is removed and single paragraph is left (childCount == 1)
+	return tableCell.childCount < 3;
```
scofalik: This will trigger a refresh when removing one of three elements, leaving with two.

tests/converters/table-cell-paragraph-post-fixer.js
jodator: I'll check, I forgot about spying on differ :man_facepalming:

src/converters/table-cell-refresh-post-fixer.js
jodator: No need to refresh.

src/converters/table-cell-refresh-post-fixer.js
jodator: There were tests for this.

src/converters/table-cell-refresh-post-fixer.js
jodator: huh, we might be more specific here

scofalik:
Okay, we talked F2F about this and the solution with `WeakMap` has a small "initialization" problem. OTOH, there are fewer cases that we really have to worry about, so I trust the proposed solution more.

We concluded that we can go with the proposed solution but I'd like to see more tests to feel safer and maybe some additional comments in the code explaining why the `if`s that we have are enough and what assumptions we took so that it can work (for example, we were talking that in the end we only need to refresh if there is a paragraph because only existing paragraph might need that -- there is an `if` in the code but it is not explained clear enough).

src/converters/table-cell-refresh-post-fixer.js
scofalik: We concluded that it might be for pasting so let's leave this `if`.

jodator:
I've rewrite this one case and added more comments on why those checks.

Also I've updated the **proper** test file - I don't know why I've changed tests for other post-fixer :man_facepalming:.

jodator:
One more finding... this text wasn't suppose to be here.

src/converters/table-cell-refresh-post-fixer.js
```diff
@@ -25,17 +25,61 @@ export default function injectTableCellRefreshPostFixer( model ) {
...
+//
+// For instance: The <span> to <p> should be renamed when adding an attribute to a <paragraph>.
+// But adding one attribute and removing another will result in false positive: the check for added attribute will see one attribute
+// on a paragraph and will falsy qualify such change as adding attribute to a paragraph without any attribute.
+//
+// @param {module:engine/model/element~Element} tableCell Table cell to check.
+// @param {String} type Type of change.
+function checkRefresh( tableCell, type ) {
+	// If all children of a table cell were removed - refresh it.
+	if ( !tableCell.childCount ) {
```
scofalik: Actually, now I wonder, do we have to do that? Didn't we have another post-fixer for that earlier? 🤔

scofalik: What I mean is that we need to auto-paragraph this cell rather than refresh it, but refreshing it might trigger auto-paragraphing (in other post-fixer). And I just wonder if this shouldn't be elsewhere (if it is needed) or removed.

jodator: yeah.. probably also

jodator: fixing

jodator: fixed

jodator: fixed

PR #216, opened 2019-09-18
t/200 Made the InsertColumnCommand work in the opposite direction when Loca…

### LINKED ISSUE #200: [RTL] Make sure that insert column buttons work right
As in https://github.com/ckeditor/ckeditor5/issues/1151#issuecomment-515064624:

> No, I mean in the table itself. When "Insert column right", a left column inserted instead and vice versa.

A follow-up of https://github.com/ckeditor/ckeditor5/issues/1151.

oleq:

Fix: Made the `InsertColumnCommand` and `MergeCellCommand` work in the opposite direction when `Locale#contentLanguageDirection` is `'rtl'`. Closes #200.

### DIFF
--- a/src/commands/insertcolumncommand.js
+++ b/src/commands/insertcolumncommand.js
@@ -70,15 +70,35 @@ export default class InsertColumnCommand extends Command {
 		const editor = this.editor;
 		const selection = editor.model.document.selection;
 		const tableUtils = editor.plugins.get( 'TableUtils' );
+		const contentLanguageDirection = editor.locale.contentLanguageDirection;
 
 		const firstPosition = selection.getFirstPosition();
 
 		const tableCell = findAncestor( 'tableCell', firstPosition );
 		const table = tableCell.parent.parent;
 
-		const { column } = tableUtils.getCellLocation( tableCell );
-		const insertAt = this.order === 'right' ? column + 1 : column;
+		let { column } = tableUtils.getCellLocation( tableCell );
+		const isOrderRight = this.order === 'right';
+		const isContentLtr = contentLanguageDirection === 'ltr';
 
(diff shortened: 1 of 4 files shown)


oleq:
I figured merge cell commands should be mirrored too. WIP.

oleq:
Now it's ready for review.

src/commands/insertcolumncommand.js
```diff
@@ -70,15 +70,35 @@ export default class InsertColumnCommand extends Command {
...
+		//                              │            order             │
+		//                              ├───────────────┰──────────────┤
+		//                              │    'right'    │    'left'    │
+		//          ┌───────────┰───────┼───────────────┼──────────────┤
+		//          │   content │ 'ltr' │   column+1    │    column    │
+		//          │  language │───────┼───────────────┼──────────────┤
+		//          │ direction │ 'rtl' │    column     │   column+1   │
+		//          └───────────┴───────┴───────────────┴──────────────┘
+		//
+		if ( ( isOrderRight && isContentLtr ) || ( !isOrderRight && !isContentLtr ) ) {
```
jodator: This gets complicated with RTL. And I wonder if we shouldn't put emphasis on the UI for dictating "left"/"right" and have "before"/"after" in the model?

This way the command will accept "before" / "after" params from the UI and will execute them in the model order (as in arrays). The UI would switch command param depending on the content to match the visual aspect of it.

jodator: In other words, the model manipulation should be simpler. It should focus on data direction as in after/before. The visual aspect of the editor, at least in here, could be done on the UI side of things.

This makes sense for me that

oleq:
A better approach in https://github.com/ckeditor/ckeditor5-table/pull/218.

PR #218, opened 2019-09-19
t/200c: Column insertion and cell merging buttons should work correctly...

### LINKED ISSUE #200: [RTL] Make sure that insert column buttons work right

oleq:

Fix: Column insertion and cell merging buttons should work correctly when the editor content is right–to–left (RTL). Closes #200.

### DIFF
--- a/src/tableui.js
+++ b/src/tableui.js
@@ -38,6 +38,8 @@ export default class TableUI extends Plugin {
 	init() {
 		const editor = this.editor;
 		const t = this.editor.t;
+		const contentLanguageDirection = editor.locale.contentLanguageDirection;
+		const isContentLtr = contentLanguageDirection === 'ltr';
 
 		editor.ui.componentFactory.add( 'insertTable', locale => {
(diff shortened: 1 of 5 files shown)


scofalik:
Remember to test this with track changes ...

oleq:
@scofalik The change is at the UI level. Could this affect TC?

jodator:
Yeah - I think that this shouldn't affect TC from the model POV (I don't know what happens there on the UI level)

oleq:
This is what it looks like in LTR now. It shouldn't be a big UI deal when it goes RTL.

image

BTW: Something's wrong with merging cells. It does not look like it does anything...

scofalik:
I was afraid that after pressing "insert column left" the column would be inserted on the right or something. 👍

PR #231, opened 2020-01-30
Fix a bug with spellchecking used in a table cell crashing the editor

Reinmar:

Fix: Fixed a bug with spellchecking or pasting via the context menu into a table cell crashing the editor. Closes ckeditor/ckeditor5#6062.

### DIFF
--- a/tests/converters/downcast.js
+++ b/tests/converters/downcast.js
@@ -26,7 +26,10 @@ function paragraphInTableCell() {
 
 				if ( viewElement && viewElement.name === 'span' ) {
 					conversionApi.mapper.unbindModelElement( tableCell );
+
+					conversionApi.writer.removeStyle( 'display', viewElement );
 					conversionApi.writer.rename( 'p', viewElement );
+
 					conversionApi.mapper.bindElements( child, viewElement );
 				}
 			}
@@ -339,7 +342,9 @@ describe( 'downcast converters', () => {
 						'<table>' +
 							'<tbody>' +
 								'<tr>' +
-									'<td class="ck-editor__editable ck-editor__nested-editable" contenteditable="true"><span></span></td>' +
+									'<td class="ck-editor__editable ck-editor__nested-editable" contenteditable="true">' +
+										'<span style="display:inline-block"></span>' +
+									'</td>' +
 								'</tr>' +
 							'</tbody>' +
 						'</table>' +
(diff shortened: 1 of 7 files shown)


tests/converters/downcast.js
```diff
@@ -26,7 +26,10 @@ function paragraphInTableCell() {
 
 				if ( viewElement && viewElement.name === 'span' ) {
 					conversionApi.mapper.unbindModelElement( tableCell );
+
+					conversionApi.writer.removeStyle( 'display', viewElement );
```
Reinmar: I found such a function defined in tests quite odd. In fact, the whole _utils/utils.js file contains so much feature code that I'm afraid that many tests here are actually testing that code, not how the real table feature work in real life. 

Wouldn't it be possible to make the tests in `downcast.js` more realistic?

Reinmar: cc @jodator

Reinmar:
I didn't add any new automated test because existing ones check whether the span has this style or not and that's all we can reliably test. Other than that, it requires manual tests.

@Mgsy, I'm fairly confident about this change on Chrome/Safari. But I didn't check features such as comments or track changes with it nor did I check Firefox.

FilipTokarski:
### [ SAFARI ]
Steps:
- add table
- paste text using context menu

Result:
table_paste_error

Note: no error in console

FilipTokarski:
As it comes to spellchecker, everything seems to be fine now in Chrome and Safari

Mgsy:
It looks good in Firefox.

tests/converters/downcast.js
jodator: Yes, I don't like this either after 2 years. Previously it was done probably to isolate conversion from other editing test. IIRC this might be done as other, less schema intensive, features were tested this way (by defining schema by hand) for conversion tests.

At this stage I'm confident that whole utils should be treated with :fire:.

To make those tests more realistic we should always use `TableEditing` for those tests.

Reinmar:
> Unfortunately this does not fix [ckeditor/ckeditor5#3243](https://github.com/ckeditor/ckeditor5/issues/3243)

OK, thanks. I removed it from the covered tickets.

tests/converters/downcast.js
Reinmar: > To make those tests more realistic we should always use TableEditing for those tests.

👍 Let's keep that in mind for the future.

jodator:
LGTM :+1:

PR #232, opened 2020-01-31
I/6112: Implemented the table properties form

oleq:

Feature: Implemented the table properties form. Closes ckeditor/ckeditor5#6112.

---

### Additional information

[**CI and the constellation**](https://github.com/ckeditor/ckeditor5/tree/i/6112)

## Requires

* [x] [ckeditor5-ui](https://github.com/ckeditor/ckeditor5-ui/pull/539) — the new property allowing to make the toolbar compact (for all alignment toolbars in both table and table cell properties forms)
* [x] [ckeditor5-theme-lark](https://github.com/ckeditor/ckeditor5-theme-lark/pull/264):
   * styles for the compact toolbar,
   * styles fro the table properties form,
   * I found out that we didn't separate wireframe and theme styles in the previous PR so this PR does that
   * styles for private components like `FormRowView` and `FormHeaderView`.

## Remarks

`TablePropertiesUI` and `TableCellPropertiesUI` are very similar (same with the views). At first, I wanted to create a base class (classes) for all of them (and/or a tool to help compose them) but then I figured this may not be the best idea because even though at this moment they are very similar, they may diverge in the future and this will be a situation with an unmaintainable base class (or set of tools).

So, long story short, I don't like this huge code duplication (in UI and View classes) but at the same time, I'm worried that extracting the common pieces will not be good in the long run.

### DIFF
--- a/src/tableproperties/tablepropertiesui.js
+++ b/src/tableproperties/tablepropertiesui.js
@@ -138,7 +138,7 @@ export default class TablePropertiesUI extends Plugin {
 			if ( !getTableWidgetAncestor( viewDocument.selection ) ) {
 				this._hideView();
 			} else if ( this._isViewVisible ) {
-				repositionContextualBalloon( editor );
+				repositionContextualBalloon( editor, 'table' );
 			}
 		} );
 
--- a/src/tableproperties/tablepropertiesui.js
+++ b/src/tableproperties/tablepropertiesui.js
@@ -1,0 +1,280 @@
+/**
+ * @license Copyright (c) 2003-2019, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+/**
+ * @module table/tableproperties/tablepropertiesui
+ */
+
+import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
+import ButtonView from '@ckeditor/ckeditor5-ui/src/button/buttonview';
+import { getTableWidgetAncestor } from '../utils';
+import clickOutsideHandler from '@ckeditor/ckeditor5-ui/src/bindings/clickoutsidehandler';
+import ContextualBalloon from '@ckeditor/ckeditor5-ui/src/panel/balloon/contextualballoon';
+import TablePropertiesView from './ui/tablepropertiesview';
+import tableProperties from './../../theme/icons/table-properties.svg';
+import {
+	repositionContextualBalloon,
+	getBalloonTablePositionData
+} from '../ui/utils';
+
+const DEFAULT_BORDER_STYLE = 'none';
+const DEFAULT_ALIGNMENT = 'center';
+
+/**
+ * The table properties UI plugin. It introduces the `'tableProperties'` button
+ * that opens a form allowing to specify visual styling of an entire table.
+ *
+ * It uses the
+ * {@link module:ui/panel/balloon/contextualballoon~ContextualBalloon contextual balloon plugin}.
+ *
+ * @extends module:core/plugin~Plugin
+ */
+export default class TablePropertiesUI extends Plugin {
+	/**
+	 * @inheritDoc
+	 */
+	static get requires() {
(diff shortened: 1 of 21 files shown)


src/tableproperties/tablepropertiesui.js
```diff
@@ -0,0 +1,280 @@
...
+		const commands = this.editor.commands;
+
+		this.view.set( {
+			borderStyle: commands.get( 'tableBorderStyle' ).value || DEFAULT_BORDER_STYLE,
+			borderColor: commands.get( 'tableBorderColor' ).value || '',
+			borderWidth: commands.get( 'tableBorderWidth' ).value || '',
+			backgroundColor: commands.get( 'tableBackgroundColor' ).value || '',
+			width: commands.get( 'tableWidth' ).value || '',
+			height: commands.get( 'tableHeight' ).value || '',
+			alignment: commands.get( 'tableAlignment' ).value || DEFAULT_ALIGNMENT,
```
oleq: I'm quite puzzled about which default alignment should we use. Center? Or left? Or depending on the directionality of the content?

Also, the topic is broader than this. ATM changing the table alignment in the UI only changes the text alignment in all cells because our (block widget) styles enforce centered tables no matter the `alignment` DOM attribute value. It's a serious UX issue.

## Option 1 

We could make table widgets align their content to the left by default (technically, make sure they do nothing) and allow users to use the table properties UI to choose the alignment (I checked this, it works). But then... what would the tables look like if the properties plugin is not loaded? They should probably be centered. So this is a weird approach.

## Option 2

We could bind the alignment in the editing to the widget container class so:
* the DOM `alignment` of the `<table>` is still applied,
* the class of the widget container ensures visually everything looks right.

This class would be present only in editing. I wonder if this would actually work well (we downcast `<figure><table>...` so there's always a wrapper).

jodator: I'll look into it more during the review - but it looks like something is wrong here - the table cells shouldn't be aligned left/right by the align table command. I didn't read the whole case though - so I might be missing something.

jodator: That's a conversion bug - I don't know why the `'alignment'` attribute is downcasted here but probably the best way is to rename table attributes :/ So here it could be `tableAlignment` instead of `alignment`. I'll confirm this with an issue.

jodator:
> At first, I wanted to create a base class (classes) for all of them (and/or a tool to help compose them) but then I figured this may not be the best idea because even though at this moment they are very similar, they may diverge in the future and this will be a situation with an unmaintainable base class (or set of tools).

I had a similar feeling about all the command classes. There are two base classes but it is not a big deal to refactor them later if needed. The code duplication at this stage is OK IMO.

src/tableproperties/ui/tablepropertiesview.js
```diff
@@ -0,0 +1,653 @@
...
+		this._focusCycler.focusFirst();
+	}
+
+	/**
+	 * Creates the header of the form with a localized label.
+	 *
+	 * @private
+	 * @returns {module:table/ui/formheaderview~FormHeaderView}
+	 */
+	_createHeaderView() {
```
jodator: Similarly to comment in TebleCellProperitesView - might be inlined.

src/tableproperties/tablepropertiesui.js
Reinmar: Shouldn't table alignment should align **the entire table** to left/right/center? Not its content.

Reinmar: OK, I can see that https://github.com/ckeditor/ckeditor5/issues/6179 clarifies that the current behaviour is a result of a bug.

jodator: @Reinmar yes - details are here: https://github.com/ckeditor/ckeditor5/issues/6179.

jodator: I'd move the conversion on this to linked issue (:point_up:) as the UI operates on commands and the bug is in editing part of the feature :/

PR #233, opened 2020-02-05
I/6179: Fix alignment property converter

jodator:

Fix: Table's `alignment` attribute converter should override `Alignement` feature converter. Closes ckeditor/ckeditor5#6179.

---

### Additional information

* The where to convert `alignment` attribute is a different topic: ckeditor/ckeditor5#6197.

### DIFF
--- a/src/converters/tableproperties.js
+++ b/src/converters/tableproperties.js
@@ -105,6 +105,10 @@ export function downcastTableAttribute( conversion, modelAttribute, styleName ) 
 		const { item, attributeNewValue } = data;
 		const { mapper, writer } = conversionApi;
 
+		if ( !conversionApi.consumable.consume( data.item, evt.name ) ) {
+			return;
+		}
+
 		const table = [ ...mapper.toViewElement( item ).getChildren() ].find( child => child.is( 'table' ) );
 
 		if ( attributeNewValue ) {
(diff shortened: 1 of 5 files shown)


jodator:
There are some changes in the Chrome 80 in comparison to Chrome 79... I've just updated to Chrome 80 and indeed this one test fail. Was green on Chrome 79.

Reinmar:
👍 for consuming the attribute – that's definitely a good change.

👎 for changing the priority of the converter. As I commented in https://github.com/ckeditor/ckeditor5/issues/6179#issuecomment-581465864 the alignment feature converter should be smarter. Semantically, the alignment feature applies to text blocks (it sets `text-alignment` so it affects the content of those elements). So, it should not apply to object elements.

In other words – IMO, instead of rising the priority of the table alignment converter we should make the alignment feature's converter apply only to non-object elements.

jodator:
> -1 for changing the priority of the converter. As I commented in [ckeditor/ckeditor5#6179 (comment)](https://github.com/ckeditor/ckeditor5/issues/6179#issuecomment-581465864) the alignment feature converter should be smarter. Semantically, the alignment feature applies to text blocks (it sets `text-alignment` so it affects the content of those elements). So, it should not apply to object elements.
> 
> In other words – IMO, instead of rising the priority of the table alignment converter we should make the alignment feature's converter apply only to non-object elements.

As in F2F conversation: re-writing the `Alignment` feature converter is not a half-hour task so we extracted this to: https://github.com/ckeditor/ckeditor5/issues/6199.

@Reinmar other than that this PR is ready to review (and merge - the only failing test fails on master also because of [failing tests on Chrome 80](https://github.com/ckeditor/ckeditor5/issues/6198).

jodator:
Invalidated by #234 - the tests remained untouched though.

PR #235, opened 2020-02-06
I/6186: downcast height/width to <figure>

jodator:

Other: The table's width/height should be downcasted to the `<figure>`. Closes ckeditor/ckeditor5#6186.

---

### Additional information

* Will conflict with #234 because of the same test utility method was upgraded.


jodator:
> Are we sure this is OK?

OH it looks bad - I forgot about that. I really don't know how this should behave after all.

It looks like a better suite for `height/width` is `<table>` but it gave soem problems in `image` so addin g `.ck-content .table table` looks OK.

oleq:
Everything else LGTM, so this is the only issue. Some thoughts @Reinmar?

Reinmar:
> Everything else LGTM, so this is the only issue. Some thoughts @Reinmar?

Looks good to me. It feels slightly weird but correct at the same time :D

PR #236, opened 2020-02-07
I/6189: Copy-paste dimiensions fields to the table cell properties view.

jodator:

Feature: Add width and height fields to table cell properties view. Closes ckeditor/ckeditor5#6189.

---

### Additional information

So the copy-paste is done and it looks like this:

Selection_150

Comparing this to the table properties view:

Selection_151

I think that it would probably be better to have them "similar"

```
Border
Background
Width x Height | padding
Text Alignment
```

this way both will look similar (I like smaller inputs for width / height).

@oleq WDYT?


oleq:
I would try fitting width and height with the same row as padding, leaving more air for the background input which can have lengthy values. I can check it during the review.

Also, the CI is failing :P

jodator:
@oleq so the feature is ready to review. The only test that is failing is
> Table cell paragraph post-fixer should wrap in paragraph $text nodes placed directly in tableCell (on table cell modification)
which is the same failing on master so :man_shrugging: I think we can merge it if only this tests fails.

jodator:
The feature is ready to review. The only test that is failing is
> Table cell paragraph post-fixer should wrap in paragraph $text nodes placed directly in tableCell (on table cell modification)

which is the same failing on master so :man_shrugging: I think we can merge it if only this test fails.

PR #239, opened 2020-02-07
i/6180: Clicking cancel in a table and a table cell property forms should not remove a newly created table

oleq:

Fix: Cancelling the table and table cell property forms should not remove a newly created table. Closes ckeditor/ckeditor5#6180.

### DIFF
--- a/src/tablecellproperties/tablecellpropertiesui.js
+++ b/src/tablecellproperties/tablecellpropertiesui.js
@@ -121,7 +121,11 @@ export default class TableCellPropertiesUI extends Plugin {
 		} );
 
 		this.listenTo( view, 'cancel', () => {
-			editor.execute( 'undo', this._undoStepBatch );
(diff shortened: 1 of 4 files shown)


src/tablecellproperties/tablecellpropertiesui.js
```diff
@@ -121,7 +121,11 @@ export default class TableCellPropertiesUI extends Plugin {
 		} );
 
 		this.listenTo( view, 'cancel', () => {
-			editor.execute( 'undo', this._undoStepBatch );
+			// https://github.com/ckeditor/ckeditor5/issues/6180
+			if ( this._undoStepBatch.operations.length ) {
```
jodator: I'm not sure I like this check but for now it is OK as it uses pulbic batch API. https://github.com/ckeditor/ckeditor5/issues/6212

PR #246, opened 2020-02-12
Lazy loading for the table dropdown

mlewand:

Other: Implemented lazy loading for the table dropdown. This will reduce editor initialization time. Closes ckeditor/ckeditor5#6193.

### DIFF
--- a/src/tableui.js
+++ b/src/tableui.js
@@ -54,21 +54,29 @@ export default class TableUI extends Plugin {
 				tooltip: true
 			} );
 
-			// Prepare custom view for dropdown's panel.
-			const insertTableView = new InsertTableView( locale );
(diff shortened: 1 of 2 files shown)


mlewand:
CI tests will fail in "Table cell paragraph post-fixer should wrap in paragraph $text nodes placed directly in tableCell (on table cell modification)" - that's caused by Chrome 80 bug. Run it on Firefox to have a green ✅ run.

Reinmar:
I'm missing a test that the view is indeed lazily created. This is – it does not exist before the dropdown is opened. Like this thing: https://github.com/ckeditor/ckeditor5-special-characters/pull/13/files#r378314398.

PR #230, opened 2020-01-29
I/6113: Base table selection plugin

jodator:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Feature: Introduce `TableSelection` plugin. Closes ckeditor/ckeditor5#6113.

---

### Additional information

- :tada: a `TableSelection` has matured since its first commit in August of 2018.
- most important things here are: `TableSelection` plugin itself and the introduced `MouseSelectionHandler` - the architecture of this feature is meant to have keyboard handler in a similar fashion
- the conversion uses the highlight mechanism known from the widget (fake selection, etc)
- if something is not working or not implemented - check the epick first if it is not reported - this PR only enables table selection with a mouse - there are no commands that would support this ATM

PR #250, opened 2020-02-19
I/6118: Copy selected table fragment.

jodator:

Feature: Copy selected table fragment. Closes ckeditor/ckeditor5#6118.

---

### Additional information

* This PR is blocked by #230 so after merging that PR the base branch must be updated (no it is `i/6118` to reflect only additional changes.
* I've introduced a `TableClipboard` plugin because I anticipate more code related strictly to copy, cut and paste scenarios. Also, the `TableSelection` plugin is responsible for handling table selection only and clipboard handling is IMO beyond its scope.
* The `"cut"` event is blocked and by handled by own issue.
* The `"paste"` event isn't handled at all as it does not break the editor (AFAICS)


jodator:
The Travis build might need to be restarted as the info is amigious and everythings is fine locally.

PR #254, opened 2020-02-25
I/6119: Clear table contents on clipboard "cut" and keyboard delete

jodator:

Feature: Clear table contents on clipboard "cut" and keyboard delete. Closes ckeditor/ckeditor5#6119. Closes ckeditor/ckeditor5#6284. Closes ckeditor/ckeditor5#6301.

---

### Additional information

For the "cut" event handler I had to override (thus duplicate) default `Clipboard` plugin behavior which was removing a single range instead of multi-range selection. This could be simplified by enhancing `model.deleteContent()` (reported here: https://github.com/ckeditor/ckeditor5/issues/6328).

### DIFF
--- a/src/tableclipboard.js
+++ b/src/tableclipboard.js
@@ -10,8 +10,6 @@
 import Plugin from '@ckeditor/ckeditor5-core/src/plugin';
 import TableSelection from './tableselection';
 import { clearTableCellsContents } from './tableselection/utils';
-import viewToPlainText from '@ckeditor/ckeditor5-clipboard/src/utils/viewtoplaintext';
-import HtmlDataProcessor from '@ckeditor/ckeditor5-engine/src/dataprocessor/htmldataprocessor';
 
 /**
  * The table clipboard integration plugin.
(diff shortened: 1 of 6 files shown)


Reinmar:
> For the "cut" event handler I had to override (thus duplicate) default `Clipboard` plugin behavior which was removing a single range instead of multi-range selection. This could be simplified by enhancing `model.deleteContent()` (reported here: [ckeditor/ckeditor5#6328](https://github.com/ckeditor/ckeditor5/issues/6328)).

I think that improving `deleteContent()` and `insertContent()` is a must. Most features use those methods so we'll need to do that for https://github.com/ckeditor/ckeditor5/issues/6301, https://github.com/ckeditor/ckeditor5/issues/6284 and https://github.com/ckeditor/ckeditor5/issues/6116. But we can live with that for a second.

jodator:
@Reinmar So I've checked the state of the `deleteContent()`. ATM (this PR) we have a safe thing that works. Upgrading the `model.deleteContent()` with directionality (https://github.com/ckeditor/ckeditor5/issues/3115) could be done in a separate ticket.

From what I can see this PR fixes also (see closes in the first post):
- enter
- typing

over selected table fragment.

it does not, however, fix the https://github.com/ckeditor/ckeditor5/issues/6346 - which is required for UNDO to work :/ Ps.: IMO separate PR/ticket.

PR #253, opened 2020-02-24
I/6150: Make table cell properties UI aware of mult-cell selection.

jodator:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Enhancement: Add multi-cell selection to table cell properties UI. Closes ckeditor/ckeditor5#6150.

---

### Additional information

This PR adds the ability to style selected table cells with basic logic for command value. I've extracted a follow for for multi-cell values handling as it looks optional and requires decisions.

@Reinmar: is this follow up OK? https://github.com/ckeditor/ckeditor5/issues/6320

PR #256, opened 2020-02-28
I/6346: TableSelection should work with external changes.

jodator:

Type: Message. Closes ckeditor/ckeditor5#6346.

---

### Additional information

* Work in progress - based on https://github.com/ckeditor/ckeditor5-table/pull/254 & https://github.com/ckeditor/ckeditor5-table/pull/253 PRs so it would need to be rebased once those two got merged.

* First naive implmeentation is to "fix" TableSelection state when there's an external change detected (focused on undo case). In such state, there will be some table cells in the selection but internally it will be blank.

* However this is somewhat in a contrary to `_clearSelectionOnExternalChange()` method.


Reinmar:
Fixed by #263.

PR #266, opened 2020-03-09
Split command is now disabled when multiple cells are selected

mlewand:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Internal: Split command is now disabled when multiple cells are selected. Closes ckeditor/ckeditor5#6128.

---

### Additional information

At first I added [a generator that returns selected cells](https://github.com/ckeditor/ckeditor5-table/blob/eaf2840514185252db4e9366ce6750feb14c08ef/src/tableselection/utils.js#L53-L81) (copied it from my another work in progress PR), but that was very similar to `getTableCellsInSelection()`.

So I have enhanced `getTableCellsInSelection()` with `expandSelection` parameter that makes it useful for my case. Also it will be possible to reuse this method for more use cases, for instance for the upcoming remove row / column commands (this is in fact where I copied the initial generator from).

PR #265, opened 2020-03-09
i/6358: Implemented the `getSelectedTableCells()` helper

oleq:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Internal: Implemented the `getSelectedTableCells()` helper based on `Range#getContainedElement()`. Code refactoring. Closes ckeditor/ckeditor5#6358.

---

### Additional information

**Note**: Based on https://github.com/ckeditor/ckeditor5/issues/6364. **It must be closed first.**

PR #268, opened 2020-03-10
Integrated multi cell selection with set column /row header commands

mlewand:

Type: Integrated multi cell selection with set column /row header commands. Closes ckeditor/ckeditor5#6127.

---

### Additional information

After this PR the header command state could get a third state - "mixed" because it's possible to make selection where only part of it contains headers:

However to simplify things, I implemented it in a way that command state is true only when all selected cells contain header. Otherwise it's `false`.

~I have based this PR on #266 as this PR adds the API that was useful for this enhancement.~ Rebased onto master, after #265 added API that I was mocking before.

### DIFF
--- a/src/commands/setheadercolumncommand.js
+++ b/src/commands/setheadercolumncommand.js
@@ -71,13 +71,15 @@ export default class SetHeaderColumnCommand extends Command {
 		const tableRow = firstCell.parent;
 		const table = tableRow.parent;
 
-		const selectionColumn = Math.max( tableUtils.getCellLocation( firstCell ).column, tableUtils.getCellLocation( lastCell ).column );
+		const [ selectedColumnMin, selectedColumnMax ] =
+			// Returned cells might not necessary be in order, so make sure to sort it.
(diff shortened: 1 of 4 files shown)


mlewand:
Rebased onto the latest `master` branch from after #265. @ckeditor/qa-team please give it a testing.

mlewand:
Just pushed one missing commit, sorry for that :see\_no\_evil: @ckeditor/qa-team

mlewand:
Merged latest master, resolved conflicts.

FilipTokarski:
I'm not sure what is the desired behaviour here, but this cell splitting looks a bit weird:
  
### Safari
- ( please note how on Safari it changes whole selection, it stays on one cell in column d, under the balloon )
table_header_safari1


### Chrome:  
table_header_chrome1

mlewand:
> I'm not sure what is the desired behaviour here, but this cell splitting looks a bit weird:

I believe that exactly the same behavior is present on `master` branch, right? Neither was I a fan of this, but here I focused on bringing support for multi select, not that much on fixing underlying table plugin logic.

mlewand:
Merged the `master` branch and resolved conflicts.

PR #269, opened 2020-03-10
i/6228: Change the default background color of the table headers

panr:

Other: Changed the default background color for the table headers to be slightly darker than table background. Closes ckeditor/ckeditor5#6228.

### DIFF
--- a/theme/table.css
+++ b/theme/table.css
@@ -31,7 +31,7 @@
 
 		& th {
 			font-weight: bold;
-			background: hsl(0, 0%, 98%);
+			background: hsla(0, 0%, 0%, 10%);
 		}
 	}
 }
--- a/theme/table.css
+++ b/theme/table.css
@@ -31,7 +31,7 @@
 
 		& th {
 			font-weight: bold;
-			background: hsla(0, 0%, 0%, 10%);
+			background: hsla(0, 0%, 0%, 5%);
 		}
 	}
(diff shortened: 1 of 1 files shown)


panr:
Ready for review.

theme/table.css
```diff
@@ -3,6 +3,10 @@
  * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
  */
 
+ :root {
```
Reinmar: Do we need a variable for this, if it's used in one place?

Reinmar: I can see that you actually used it in https://github.com/ckeditor/ckeditor5-theme-lark/pull/271/files. Which IMO is wrong because you named it `th-background` and yet you use it for focused cell background. The variable has a meaning and you should use it according to its meaning.

Anyway, I don't think we need a variable for that. The ticket was about switching to an hsla color. Refactor should be justified. So the question is – why did you refactor it? Do we need to do that? If no, let's focus on the color itself.

Reinmar: > you named it `th-background` and yet you use it for focused cell background

Actually, it's the other way around – you called it "focused th background" and you use it for normal ths :P

panr: You're right, sorry about that 👍

panr:
Ready again.

PR #271, opened 2020-03-11
i/6368: Merge left and right commands should be always enabled if execution does not cross the heading column boundary

oleq:

Fix: Merge left and right commands should be always enabled if the execution does not cross the heading column boundary. Closes ckeditor/ckeditor5#6368.


oleq:
The actual fix is here https://github.com/ckeditor/ckeditor5-table/pull/271/files#diff-5c5e41ec6bf80a04107db4e13df67d01R184. The rest is code refactoring.

PR #261, opened 2020-03-03
Introduced support for removing multiple rows / columns using multi-cell selection

mlewand:
### Suggested merge commit message ([convention](https://github.com/ckeditor/ckeditor5-design/wiki/Git-commit-message-convention))

Type: Introduced support for removing multiple rows or cells when multiple columns are selected selection. Closes ckeditor/ckeditor5#6126.

---

### Additional information

Currently there's added row removal support.

*   There's one disturbing fact with removing a selection containing mixed header and regular rows. Unit tests are passing, while real editor usage throws an exception. It shows that unit tests approach could be more holistic (we're testing only model ATM AFAICT).
*   There's one known wrong handling of rowspanned cells.

There's a followup: ckeditor/ckeditor5#6370 related to: https://github.com/ckeditor/ckeditor5-table/blob/23d3a40efa716438bba317be08e73210cd42c5b2/tests/commands/removerowcommand.js#L130-L151

PR #272, opened 2020-03-11
Fixed a case when removing last row would crash the editor

mlewand:

Type: Fixed a case when removing last row would crash the editor. Closes ckeditor/ckeditor5#6370.


mlewand:
Closing the PR, I'll reopen it with the based on the master branch, as it's not so dependent on #261.

PR #276, opened 2020-03-11
Range order should not affect how various algorithms work

Reinmar:

Internal: The order of ranges in the selection should not affect things like copying a table fragment. Closes ckeditor/ckeditor5#6411.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -81,7 +81,7 @@ export function getTableWidgetAncestor( selection ) {
 export function getSelectedTableCells( selection ) {
 	const cells = [];
 
-	for ( const range of selection.getRanges() ) {
+	for ( const range of sortRanges( selection.getRanges() ) ) {
 		const element = range.getContainedElement();
 
 		if ( element && element.is( 'tableCell' ) ) {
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -81,7 +81,7 @@ export function getTableWidgetAncestor( selection ) {
 export function getSelectedTableCells( selection ) {
 	const cells = [];
 
-	for ( const range of selection.getRanges() ) {
+	for ( const range of sortRanges( selection.getRanges() ) ) {
```
oleq: There are 2 more places that would benefit from a public helper:
1. https://github.com/ckeditor/ckeditor5-restricted-editing/blob/master/src/restrictededitingmodenavigationcommand.js#L110-L116
2. https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/range.js#L933-L935

WDYT?

Reinmar: Hard to say as sorting ranges can be done in more ways (not only by using the start position).

I'll report a followup though as I realised that e.g. the `isEqual()` check in my code is not necessary.

oleq: Yeah, `isBefore/After` does the trick.

Reinmar: ckeditor/ckeditor5#6414 done

PR #277, opened 2020-03-12
Fixed a case where a table with empty row in model would break remove row command

mlewand:

Internal: Fixed a case where a table with empty row in model would break remove row command. Closes ckeditor/ckeditor5#6422.

---

### Additional information

Detailed issue explanation can be found here: https://github.com/ckeditor/ckeditor5/issues/6422#issuecomment-598155415

The source of this issue is ckeditor/ckeditor5#3280 but it's a bug that is there for a longer time. This adds additional checks in tableutils, so that broken model like that won't cause issues like that in the future.

### DIFF
--- a/src/tableutils.js
+++ b/src/tableutils.js
@@ -561,7 +561,8 @@ export default class TableUtils extends Plugin {
 	 */
 	getRows( table ) {
 		return [ ...table.getChildren() ].reduce( ( rows, row ) => {
-			const currentRowCount = parseInt( row.getChild( 0 ).getAttribute( 'rowspan' ) || 1 );
+			const firstCell = row.getChild( 0 );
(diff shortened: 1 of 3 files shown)


Reinmar:
cc @ckeditor/qa-team could you verify this change?

Reinmar:
IMO, the fix may need to be somewhere else – namely, ckeditor/ckeditor5#3280 should be resolved. I'd like to know why did you choose to solve it this way?

Reinmar:
IMO, the fix may need to be somewhere else – namely, ckeditor/ckeditor5#3280 should be resolved. I'd like to know why did you choose to solve it this way?

Mgsy:
Changes work fine. The problem has been fixed and I haven't found any regression.

mlewand:
@Reinmar you can find the answer in https://github.com/ckeditor/ckeditor5/issues/6422#issuecomment-598224993

PR #283, opened 2020-03-23
Fixed horizontal alignment of table cell in RTL content.

niegowski:

Fix: Horizontal alignment of table cell in RTL content. Closes ckeditor/ckeditor5#6371.

---

### Additional information

Justify icon looks odd for RTL languages - probably should be mirrored?

### DIFF
--- a/src/tablecellproperties/tablecellpropertiesediting.js
+++ b/src/tablecellproperties/tablecellpropertiesediting.js
@@ -69,6 +69,7 @@ export default class TableCellPropertiesEditing extends Plugin {
 		const editor = this.editor;
 		const schema = editor.model.schema;
 		const conversion = editor.conversion;
+		const locale = editor.locale;
 
 		editor.data.addStyleProcessorRules( addBorderRules );
 		enableBorderProperties( schema, conversion );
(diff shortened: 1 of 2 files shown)


Reinmar:
I tested this manually and I can't see a difference with master. But perhaps I'm not getting what the issue was. Or there's still something wrong. E.g., shouldn't table cells be right aligned (also visually – in the editor) in a newly created cells. In fact, it's also left aligned in the entire editor:



I think that the RTL sample should look more like this:



So it seems that there's something else wrong with the table's RTL manual test itself. Am I right, @oleq?

niegowski:
@Reinmar it's because content language in manual test for RTL is set to 'en' - https://github.com/ckeditor/ckeditor5-table/blob/master/tests/manual/rtl.js#L19

oleq:
...which looks like a [forgotten change made for some development purpose.](https://github.com/ckeditor/ckeditor5-table/commit/e9810125216ad903c975bbb2f2bf70513d51ccd9) @panr?

panr:
> ...which looks like a forgotten change made for some development purpose. @panr?

If `ui: ar`  should always correspond with `content: ar`, then yes, it's a bug here.

PR #286, opened 2020-03-25
I/6406: Fix logic behind removing multiple columns or rows in a row

jodator:

Fix: Recalculate table geometry on each row or column removal. Closes ckeditor/ckeditor5#6406.

---

### Additional information

I've noticed that the `tableMap` was passed but it is static. You must get the table map after each row/column removal for loop operations otherwise the remove row/column algorithm would work on obsolete data.


Reinmar:
I'm not going to lie to you – this is not something I want to verify by analysing the implementation 🤣I've briefly scanned the code but that's it.

cc @ckeditor/qa-team, could you go through the various cases that were failing recently and perhaps some more?

jodator:
> I'm not going to lie to you – this is not something I want to verify by analysing the implementation

Sure - the change was minimal actually - regenerate `tableMap` by calling `new TableWalker()` after each row/column removal. Otherwise the algorithm was analysing obsolete data.

FilipTokarski:
Sometimes when the table has complex geometry and you remove a row ( with multiselect or not ), cells are changing their position in table:
  
26_table1
  
26_table2

jodator:
> Sometimes when the table has complex geometry and you remove a row ( with multiselect or not )

If this happens for one row selection then it looks like a bug in the original remove row algorithm. So if nothing similar happens I'm confident that this change is OK. I'd move above case as a general table issue because it looks like an edge case as those spans are tough.

BTW, thanks for coloured samples - it is so easy to spot what's happening in the case :)

Mgsy:
We've tested it and changes look good 👍

src/commands/removerowcommand.js
```diff
@@ -33,15 +33,16 @@ export default class RemoveRowCommand extends Command {
...
-			const table = firstCell.parent.parent;
+			const table = findAncestor( 'table', firstCell );
 			const tableRowCount = this.editor.plugins.get( 'TableUtils' ).getRows( table );
+			const lastRowIndex = tableRowCount - 1;
 
-			const tableMap = [ ...new TableWalker( table ) ];
-			const rowIndexes = tableMap.filter( entry => selectedCells.includes( entry.cell ) ).map( el => el.row );
-			const minRowIndex = rowIndexes[ 0 ];
-			const maxRowIndex = rowIndexes[ rowIndexes.length - 1 ];
+			const removedRowIndexes = getRowIndexes( selectedCells );
```
mlewand: These cells are not yet removed, these are selected cells that are about to be removed.

PR #289, opened 2020-03-27
Fix: Inserting table column at first col-row spanned cell should work properly.

niegowski:

Fix: Inserting table column at the first col-row spanned cell should work properly. Closes ckeditor/ckeditor5#5882.

---

### DIFF
--- a/src/tableutils.js
+++ b/src/tableutils.js
@@ -228,7 +228,7 @@ export default class TableUtils extends Plugin {
 				const rowspan = parseInt( cell.getAttribute( 'rowspan' ) || 1 );
 				const colspan = parseInt( cell.getAttribute( 'colspan' ) || 1 );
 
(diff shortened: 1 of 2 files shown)


src/tableutils.js
```diff
@@ -228,7 +228,7 @@ export default class TableUtils extends Plugin {
 				const rowspan = parseInt( cell.getAttribute( 'rowspan' ) || 1 );
 				const colspan = parseInt( cell.getAttribute( 'colspan' ) || 1 );
 
-				if ( cell.index !== insertAt && colspan > 1 ) {
+				if ( cellIndex !== insertAt && colspan > 1 ) {
```
jodator: Nice catch!

Reinmar: 🎼🎼🎼🎼

PR #290, opened 2020-03-27
I/6123: The MergeTableCells command

jodator:

Feature: Introduce `MergeTableCellsCommand`. Closes ckeditor/ckeditor5#6123.

---

### Additional information

This should work OK now :)


jodator:
Yo, @ckeditor/qa-team ! This needs some deep checks. So far I found (and fixed) some issues related to the direction of a selection (ie I had a bug when selecting from bottom right to top left).

You should not be able to merge cells from heading section (rows) with other cells.

Other than that everything should be possible, clean table should be preserved (no extra empty cells). Content (images, paragraphs) should be preserved, undo should work, etc.

The button is available for now in the merge cells menu.

Mgsy:
We've tested it and everything looks good 👌

oleq:
Are you sure this makes sense?

image

and some weird result

image

jodator:
> Are you sure this makes sense?

Oh, forgot about this. I'll create a follow for this case.

The second thing (merging more then selected) - is this a bug or different screenshot?

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+
+		if ( overlapsRemovedRow ) {
+			updateNumericAttribute( 'rowspan', rowspan - 1, cell, writer );
+		}
+	}
+
+	writer.remove( row );
+}
+
+// Merges two table cells - will ensure that after merging cells with empty paragraph the result table cell will only have one paragraph.
```
oleq, line 92:
```suggestion
// Merges two table cells - will ensure that after merging cells with empty paragraphs the result table cell will only have one paragraph.
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+// If one of the merged table cell is empty the merged table cell will have contents of the non-empty table cell.
```
oleq, line 93:
```suggestion
// If one of the merged table cells is empty, the merged table cell will have contents of the non-empty table cell.
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+// If both are empty the merged table cell will have only one empty paragraph.
```
oleq, line 94:
```suggestion
// If both are empty, the merged table cell will have only one empty paragraph.
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+		}
+
+		writer.move( writer.createRangeIn( cellBeingMerged ), writer.createPositionAt( targetCell, 'end' ) );
+	}
+
+	// Remove merged table cell.
+	writer.remove( cellBeingMerged );
+}
+
+// Checks if passed table cell contains empty paragraph.
```
oleq, line 112:
```suggestion
// Checks if the passed table cell contains an empty paragraph.
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+//
+// @param {module:engine/model/element~Element} tableCell
+// @returns {Boolean}
+function isEmpty( tableCell ) {
+	return tableCell.childCount == 1 && tableCell.getChild( 0 ).is( 'paragraph' ) && tableCell.getChild( 0 ).isEmpty;
+}
+
+// Check if selection contains mergeable cells.
```
oleq, line 120:
```suggestion
// Checks if the selection contains mergeable cells.
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+//
+// In a table below:
+//
+//   +---+---+---+---+
```
oleq, line 124: Take a look at https://en.wikipedia.org/wiki/Box-drawing_character. They rock :p

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+//   | a | b | c | d |
+//   +---+---+---+   +
+//   | e     | f |   |
+//   +       +---+---+
+//   |       | g | h |
+//   +---+---+---+---+
+//
+// Valid selections are those which creates a solid rectangle (without gaps), such as:
```
oleq, line 132:
```suggestion
// Valid selections are these which create a solid rectangle (without gaps), such as:
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+//   - a, b (two horizontal cells)
+//   - c, f (two vertical cells)
+//   - a, b, e (cell "e" spans over four cells)
+//   - c, d, f (cell d spans over cell in row below)
```
oleq, line 136:
```suggestion
//   - c, d, f (cell d spans over a cell in the row below)
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+//
+// While invalid selection would be:
```
oleq, line 138:
```suggestion
// While an invalid selection would be:
```

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+	// Collapsed selection or selection only one range can't contain mergeable table cells.
+	if ( selection.isCollapsed || selection.rangeCount < 2 ) {
+		return false;
+	}
+
+	const selectedTableCells = getSelectionAffectedTableCells( selection );
+
+	if ( !areCellInTheSameTableSection( selectedTableCells ) ) {
+		return false;
+	}
```
oleq, line 155: Wouldn't simple 
```js
const selectedTableCells = getSelectedTableCells( ... );

if ( selectedTableCells.length < 2 ) {
    return;
} 
```
be enough?

src/commands/mergecellscommand.js
jodator: As a test I've used them in this example. I don't like the gap between rows but it looks better :)

jodator: As a test I've used them in this example. I don't like the gap between rows but it looks better :)

src/commands/mergecellscommand.js
jodator: Would be enough but checking selection without calculating stuff is IMO a bit faster. The `getSelectedTableCells()` will do all sort of things and here I didn't wanted to calculate all those things.

Probably.

This code is old and I've done some refactoring here already. So WDYT about leaving this as is?

src/commands/mergecellscommand.js
```diff
@@ -0,0 +1,246 @@
...
+			// Update target cell dimensions.
+			const { mergeWidth, mergeHeight } = getMergeDimensions( firstTableCell, selectedTableCells, tableUtils );
+			updateNumericAttribute( 'colspan', mergeWidth, firstTableCell, writer );
+			updateNumericAttribute( 'rowspan', mergeHeight, firstTableCell, writer );
+
+			for ( const tableCell of selectedTableCells ) {
+				const tableRow = tableCell.parent;
+				mergeTableCells( tableCell, firstTableCell, writer );
+				removeRowIfEmpty( tableRow, writer );
+			}
```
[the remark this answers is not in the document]
jodator: I'll look into it for a quick win. If I don't manage to find any let's move it to a follow-up.

src/commands/mergecellscommand.js
oleq: 1. `getSelectionAffectedTableCells()` executes `getSelectedTableCells()` anyway (see the code). 
2. But `getSelectionAffectedTableCells()` also returns cells that have  a selection **inside** of them (see `getTableCellsContainingSelection()`), which, if not for the initial condition (`selection.isCollapsed || selection.rangeCount < 2)`, would yield invalid cells from the perspective of multi-cell merge. 
3. So considering ☝️ , there's no point in using `getSelectionAffectedTableCells`. `getSelectedTableCells` is enough.

--- 

1. As for the initial condition: I understand that it spares us some unnecessary execution of `getSelectedTableCells()`.
2. But the condition is weak. For instance, you can have a selection with `selection.rangeCount === 3` in FF, each range spanning over some text in a cell, and that condition will pass. 
   2.1. It will also pass if anyone create such selection from code.
   2.2. (referring to the previous section of this comment): in conjunction with `getSelectionAffectedTableCells()` (instead of `getSelectedTableCells()`) this will cause bugs
4. I think it would be better if this condition was moved to `getSelectedTableCells()` as an initial filter to speed things up everywhere, not only here.

---

To sum up:
* the condition should land at the beginning of `getSelectedTableCells()` (return an empty array if `false`) as a speed benefit for all the sub-features,
* then it's enough to check if `getSelectedTableCells()` has length greater than 1.

src/commands/mergecellscommand.js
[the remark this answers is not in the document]
jodator: Quick wins are in API and it doesn't look good. TL;DR: follow-up and performance research. 

Fiddling with the merge cells algorithm didn't get me any noticeable performance boost (and AFAICS was adding small time). I've tried to move thing in the first step and them remove other things (cells, rows) in the next step. That didn't helped.

What _did_ help was checking what took the most time and try to optimize those.

Original time: `_undo` 650ms:



The `is` time is from [`RootElement.is()`](https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/rootelement.js#L82-L88) - there's a `.replace` called on a string. I've removed that:

After removing this: `_undo` took 566 ms:



And self time of `is` dropped by ~100ms. The second one was in `Position` constructor. [Here](https://github.com/ckeditor/ckeditor5-engine/blob/master/src/model/position.js#L85) we do `.concat()` of arrays. DevTools marked that line as one that took long time. I've changed that to using spraed operator (wild guess): `path = [ ...root.getPath(), ...path ];` and again perf boost, `_undo` took 465 ms:



and self-time of `Position` constructor dropped by ~100ms.

This is something in line with our performance research but I worry that here we just do a lot of stuff. Many operations that requires re-calculating positions, ranges and doing _many_ transformations.

src/commands/mergecellscommand.js
jodator: Sure fair enough. I've forgot about Firefox.

src/commands/mergecellscommand.js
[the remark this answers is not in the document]
jodator: cc @Reinmar, @mlewand :point\_up:

src/commands/mergecellscommand.js
jodator: I've updated the code.

jodator: I've updated the code.

src/commands/mergecellscommand.js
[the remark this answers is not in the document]
Reinmar: Congrats! Great findings. Completely valid changes to me. Both `Position` and `is()` appear in many cases that I've been investigating so those changes are totally worth doing. I'd be also interested in verifying how they affect some of our performance tests (see the manual tests in ckeditor5). Could you do the comparison?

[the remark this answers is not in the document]
jodator: I'll create a separate ticket for them as I did not implement this yet.

PR #291, opened 2020-03-27
Position of table cell properties balloon should be in relation to multiple selected cells.

niegowski:

Other: Position of table cell properties balloon should be in relation to multiple selected cells. Closes ckeditor/ckeditor5#6357.

---

### DIFF
--- a/src/ui/utils.js
+++ b/src/ui/utils.js
@@ -84,26 +84,26 @@ export function getBalloonTablePositionData( editor ) {
 export function getBalloonCellPositionData( editor ) {
 	const mapper = editor.editing.mapper;
 	const domConverter = editor.editing.view.domConverter;
-	const ranges = Array.from( editor.model.document.selection.getRanges() );
+	const selection = editor.model.document.selection;
 
-	if ( ranges.length > 1 ) {
+	if ( selection.rangeCount > 1 ) {
 		return {
-			target: () => createBoundingRect( ranges, modelRange => {
+			target: () => createBoundingRect( selection.getRanges(), modelRange => {
 				const modelTableCell = getTableCellAtPosition( modelRange.start );
 				const viewTableCell = mapper.toViewElement( modelTableCell );
 				return new Rect( domConverter.viewToDom( viewTableCell ) );
 			} ),
 			positions: BALLOON_POSITIONS
 		};
-	} else {
-		const modelTableCell = getTableCellAtPosition( ranges[ 0 ].start );
-		const viewTableCell = mapper.toViewElement( modelTableCell );
-
-		return {
-			target: domConverter.viewToDom( viewTableCell ),
-			positions: BALLOON_POSITIONS
-		};
 	}
+
+	const modelTableCell = getTableCellAtPosition( selection.getFirstPosition() );
+	const viewTableCell = mapper.toViewElement( modelTableCell );
+
+	return {
+		target: domConverter.viewToDom( viewTableCell ),
+		positions: BALLOON_POSITIONS
+	};
 }
 
 /**
--- a/src/ui/utils.js
+++ b/src/ui/utils.js
@@ -497,9 +497,9 @@ function getTableCellAtPosition( position ) {
 // Returns bounding rect for list of rects.
 //
 // @param {Array.<module:utils/dom/rect~Rect>|Array.<*>} list List of `Rect`s or any list to map by `mapFn`.
-// @param {Function} [mapFn] Mapping function for list elements.
+// @param {Function} mapFn Mapping function for list elements.
 // @returns {module:utils/dom/rect~Rect}
-function createBoundingRect( list, mapFn = null ) {
+function createBoundingRect( list, mapFn ) {
 	const rectData = {
 		left: Number.POSITIVE_INFINITY,
 		top: Number.POSITIVE_INFINITY,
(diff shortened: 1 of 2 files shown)


src/ui/utils.js
```diff
@@ -478,3 +493,31 @@ function getTableCellAtPosition( position ) {
...
+function createBoundingRect( list, mapFn = null ) {
+	const rectData = {
+		left: Number.POSITIVE_INFINITY,
+		top: Number.POSITIVE_INFINITY,
+		right: Number.NEGATIVE_INFINITY,
+		bottom: Number.NEGATIVE_INFINITY
+	};
+
+	for ( const item of list ) {
+		const rect = mapFn ? mapFn( item ) : item;
```
jodator: Coverage dropped and this else branch isn't covered.

src/ui/utils.js
```diff
@@ -478,3 +493,31 @@ function getTableCellAtPosition( position ) {
 
 	return isTableCellSelected ? position.nodeAfter : findAncestor( 'tableCell', position );
 }
+
+// Returns bounding rect for list of rects.
+//
+// @param {Array.<module:utils/dom/rect~Rect>|Array.<*>} list List of `Rect`s or any list to map by `mapFn`.
+// @param {Function} [mapFn] Mapping function for list elements.
+// @returns {module:utils/dom/rect~Rect}
+function createBoundingRect( list, mapFn = null ) {
```
jodator: The `null` default value isn't tested (coverage dropped):



The other branch is below.

src/ui/utils.js
jodator: Please check if tests needed update or if the branches are really needed.

src/ui/utils.js
```diff
@@ -478,3 +493,31 @@ function getTableCellAtPosition( position ) {
 
 	return isTableCellSelected ? position.nodeAfter : findAncestor( 'tableCell', position );
 }
+
+// Returns bounding rect for list of rects.
+//
+// @param {Array.<module:utils/dom/rect~Rect>|Array.<*>} list List of `Rect`s or any list to map by `mapFn`.
+// @param {Function} mapFn Mapping function for list elements.
+// @returns {module:utils/dom/rect~Rect}
+function createBoundingRect( list, mapFn ) {
```
jodator: This method (by name) looks _potentially_ interesting for `Rect` class but not now. As a general rule we tend to not create utils used in one place.

However, this made me think that this method is constructed in a way that cannot be moved easily from this place as it operates on `mapFn` to create rects.

For now, I think it would be cleaner to make this method operate on rects only (thus doing one thing only - as opposed to doing 2 now).

So the desired workflow with this method would look like this:

```js
const ranges = Array.from( editor.model.document.selection.getRanges() );
// getRectForTableCell is current mapFn - might be inlined as well
const rects = ranges.map( getRectForTableCell );

return {
    target: mergeRects( rects ),
    positions: BALLOON_POSITIONS
}
```

src/ui/utils.js
```diff
@@ -81,14 +82,28 @@ export function getBalloonTablePositionData( editor ) {
...
+	} else {
+		const modelTableCell = getTableCellAtPosition( ranges[ 0 ].start );
+		const viewTableCell = mapper.toViewElement( modelTableCell );
 
-	return {
-		target: editor.editing.view.domConverter.viewToDom( viewTableCell ),
-		positions: BALLOON_POSITIONS
-	};
+		return {
+			target: domConverter.viewToDom( viewTableCell ),
```
jodator: Might use `getRectForTableCell` (`mapFn`) here AFAICS (to check) - two lines less :D

src/ui/utils.js
```diff
@@ -81,14 +82,28 @@ export function getBalloonTablePositionData( editor ) {
...
+		return {
+			target: () => createBoundingRect( ranges, modelRange => {
+				const modelTableCell = getTableCellAtPosition( modelRange.start );
+				const viewTableCell = mapper.toViewElement( modelTableCell );
+				return new Rect( domConverter.viewToDom( viewTableCell ) );
+			} ),
+			positions: BALLOON_POSITIONS
+		};
+	} else {
+		const modelTableCell = getTableCellAtPosition( ranges[ 0 ].start );
```
jodator: `ranges[ 0 ].start` => [`editor.document.selection.getFirstPosition()`](https://ckeditor.com/docs/ckeditor5/latest/api/module_engine_model_documentselection-DocumentSelection.html#function-getFirstPosition)

src/ui/utils.js
```diff
@@ -478,3 +493,31 @@ function getTableCellAtPosition( position ) {
...
 	return isTableCellSelected ? position.nodeAfter : findAncestor( 'tableCell', position );
 }
+
+// Returns bounding rect for list of rects.
+//
+// @param {Array.<module:utils/dom/rect~Rect>|Array.<*>} list List of `Rect`s or any list to map by `mapFn`.
+// @param {Function} mapFn Mapping function for list elements.
+// @returns {module:utils/dom/rect~Rect}
+function createBoundingRect( list, mapFn ) {
+	const rectData = {
```
jodator: that's minor, but could you refactor this to just `rect` - I can see that other code for rects is using this pattern for calculations and it will be easier to search for rect calculations in the whole codebase.

niegowski: it would conflict with `rect` variable in loop

src/ui/utils.js
niegowski: As we discussed on a call, we can leave this in current form. This avoids multiple iterations over same list.

src/ui/utils.js
niegowski: Single range selection returns DOM element instead of function returning `Rect`.

PR #295, opened 2020-04-01
Added SelectRowCommand and SelectColumnCommand.

niegowski:

Feature: "Select entire column/row" added to table row/column dropdowns. Closes ckeditor/ckeditor5#6500.

---


src/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,65 @@
...
+import TableWalker from '../tablewalker';
+import { findAncestor } from './utils';
+import { getSelectionAffectedTableCells } from '../utils';
+
+/**
+ * The select column command.
+ *
+ * The command is registered by {@link module:table/tableediting~TableEditing} as the `'selectTableColumn'` editor command.
+ *
+ * To select the column containing the selected cell, execute the command:
```
jodator, line 21: Could you update the docs to reflect that it can select multiple columns? (also in the row command)

jodator, line 21: Could you update the docs to reflect that it can select multiple columns? (also in the row command)

src/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,65 @@
...
+		const endLocation = tableUtils.getCellLocation( lastCell );
+
+		const startColumn = Math.min( startLocation.column, endLocation.column );
+		const endColumn = Math.max( startLocation.column, endLocation.column );
+
+		const cellsToSelect = [];
+
+		for ( const cellInfo of new TableWalker( findAncestor( 'table', firstCell ) ) ) {
+			if ( cellInfo.column >= startColumn && cellInfo.column <= endColumn ) {
+				cellsToSelect.push( cellInfo.cell );
```
jodator, line 57: I wonder if doing one less loop wouldn't be better. Here you can use `model.createRangeOn()` and have `rangesToSelect` array.

jodator: However I don't see any performance gain on 20 columns so maybe it is a bit a stretch here.

src/commands/selectrowcommand.js
```diff
@@ -0,0 +1,56 @@
...
+	 */
+	execute() {
+		const model = this.editor.model;
+		const referenceCells = getSelectionAffectedTableCells( model.document.selection );
+		const rowIndexes = getRowIndexes( referenceCells );
+
+		const table = findAncestor( 'table', referenceCells[ 0 ] );
+		const cellsToSelect = [];
+
+		for ( const cellInfo of new TableWalker( table, { startRow: rowIndexes.first, endRow: rowIndexes.last } ) ) {
```
jodator, line 48: I'd go with iterating over `table.getChildren()` or event `for` loop and retrieving rows by index from table. Then for every row select all children (table cells). The `TableWalker` is cool to deal with columns and spans when needed but here we have simpler solution as operations on row indexes is straightforward (those maps 1:1 even with rowspans). As opposite to working with columns indexes where there's no 1:1 mapping to columns.

tests/_utils/utils.js
```diff
@@ -338,8 +338,6 @@ export function assertTableCellStyle( editor, tableCellStyle ) {
  *		] );
  *
  * The above call will check if table has second column selected (assuming no spans).
- *
```
jodator, line 341: This note is important here. I think that it could be improved but there's must be info about this IMO.

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+
+		it( 'should be false if selection is outside a table', () => {
+			setData( model, '<paragraph>11[]</paragraph>' );
+
+			expect( command.isEnabled ).to.be.false;
+		} );
+	} );
+
+	describe( 'execute()', () => {
+		it( 'should select a given column', () => {
```
jodator, line 67:
```suggestion
		it( 'should select a column of a table cell with a collapsed selection', () => {
```

The column is not passed as a parameter (I'd assume that by reading "given").

jodator: Could you also update other tests descriptions with ("given") to avoid confusion?

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+
+			command.execute();
+
+			assertSelectedCells( model, [
+				[ 0, 1 ],
+				[ 0, 1 ]
+			] );
+		} );
+
+		it( 'should select col-spanned cells', () => {
```
jodator, line 129: Could you add test also with selected:

1.  `40` and `41`
2.  `10`
3.  `31` and `33`

This might go as own `describe` block as this table is quite nice for `colspan` testing.

Although we have 100% CC we must write more tests for spanned cells as this is often neglected and can cause troubles when other will refactor code. One test for `colspan` isn't enough.

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+				[ { colspan: 2, contents: '20' }, '22', '23' ],
+				[ '30', { colspan: 2, contents: '31' }, '33' ],
+				[ '40', '41[]', '42', '43' ]
+			] ) );
+
+			command.execute();
+
+			/* eslint-disable no-multi-spaces */
+			assertSelectedCells( model, [
+				[ 0          ],
```
jodator, line 153: Nice notation :) It is more clear that way.

tests/commands/selectrowcommand.js
```diff
@@ -0,0 +1,338 @@
...
+				defaultConversion( editor.conversion );
+			} );
+	} );
+
+	afterEach( () => {
+		return editor.destroy();
+	} );
+
+	describe( 'isEnabled', () => {
+		it( 'should be true if selection is inside table cell', () => {
```
jodator, line 35:
```suggestion
		it( 'should be true if the selection is inside table cell', () => {
```

tests/commands/selectrowcommand.js
```diff
@@ -0,0 +1,338 @@
...
+			setData( model, modelTable( [
+				[ '00[]', '01' ],
+				[ '10', '11' ]
+			] ) );
+
+			expect( command.isEnabled ).to.be.true;
+		} );
+
+		it( 'should be true if selection contains multiple cells', () => {
```
jodator, line 44:
```suggestion
		it( 'should be true if the selection contains multiple cells', () => {
```

tests/commands/selectrowcommand.js
```diff
@@ -0,0 +1,338 @@
...
+
+			tableSelection._setCellSelection(
+				modelRoot.getNodeByPath( [ 0, 0, 0 ] ),
+				modelRoot.getNodeByPath( [ 0, 0, 1 ] )
+			);
+
+			expect( command.isEnabled ).to.be.true;
+		} );
+
+		it( 'should be false if selection is outside a table', () => {
```
jodator, line 59:
```suggestion
		it( 'should be false if the selection is outside a table', () => {
```

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+				defaultConversion( editor.conversion );
+			} );
+	} );
+
+	afterEach( () => {
+		return editor.destroy();
+	} );
+
+	describe( 'isEnabled', () => {
+		it( 'should be true if selection is inside table cell', () => {
```
jodator, line 35:
```suggestion
		it( 'should be true if the selection is inside table cell', () => {
```

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+			setData( model, modelTable( [
+				[ '00[]', '01' ],
+				[ '10', '11' ]
+			] ) );
+
+			expect( command.isEnabled ).to.be.true;
+		} );
+
+		it( 'should be true if selection contains multiple cells', () => {
```
jodator, line 44:
```suggestion
		it( 'should be true if the selection contains multiple cells', () => {
```

tests/commands/selectcolumncommand.js
```diff
@@ -0,0 +1,332 @@
...
+
+			tableSelection._setCellSelection(
+				modelRoot.getNodeByPath( [ 0, 0, 0 ] ),
+				modelRoot.getNodeByPath( [ 0, 0, 1 ] )
+			);
+
+			expect( command.isEnabled ).to.be.true;
+		} );
+
+		it( 'should be false if selection is outside a table', () => {
```
jodator, line 59:
```suggestion
		it( 'should be false if the selection is outside a table', () => {
```

tests/commands/selectrowcommand.js
```diff
@@ -0,0 +1,338 @@
...
+			// +----+----+----+----+----+
+			// | 00 | 01 | 02 | 03 | 04 |
+			// +    +    +    +----+----+
+			// |    |    |    | 13 | 14 |
+			// +    +    +----+    +----+
+			// |    |    | 22 |    | 24 |
+			// +    +----+----+----+----+
+			// |    | 31 | 32 | 33 | 34 |
+			// +----+----+----+----+----+
+			setData( model, modelTable( [
```
jodator, line 139: Similarly to the comment in column comand - please add more cases with `rowspan`.

lang/contexts.json
```diff
@@ -4,11 +4,13 @@
 	"Insert column left": "Label for the insert table column to the left of the current one button.",
 	"Insert column right": "Label for the insert table column to the right of the current one button.",
 	"Delete column": "Label for the delete table column button.",
+	"Select entire column": "Label for the select table entire column button.",
```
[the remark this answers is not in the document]
jodator: @Reinmar yay/no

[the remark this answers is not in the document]
Reinmar: Yes, I think it can be "select column".

In fact, it should be "select column" or "select columns" depending on the selection, but the same applies to the other labels (e.g. delete column) and we ignored this for the moment.

PR #299, opened 2020-04-07
i/6486: Implemented a button that merges table cells directly from the table toolbar

oleq:

Feature: Implemented a button that merges table cells directly from the table toolbar. Closes ckeditor/ckeditor5#6486.


oleq:
Some follow-up https://github.com/ckeditor/ckeditor5/issues/6567.

PR #301, opened 2020-04-08
I/6437: Fix - the table breaks after removing a last header row

jodator:

Fix: Removing the last header row no longer breaks the table in the editing view. Closes ckeditor/ckeditor5#6437.

---

### Additional information

~The actual fix is in 242a2e0 commit. It turned out that the tests weren't written in the same way as row removing is implemented (df80eec).~ Removing row from a header section involves changing `headingRows` attribute on a table.

The rest of the changes in tests was done due to ckeditor/ckeditor5#6574.

### DIFF
--- a/src/commands/removerowcommand.js
+++ b/src/commands/removerowcommand.js
@@ -60,7 +60,10 @@ export default class RemoveRowCommand extends Command {
 
 		const columnIndexToFocus = this.editor.plugins.get( 'TableUtils' ).getCellLocation( firstCell ).column;
 
-		model.change( writer => {
+		// Use single batch to modify table in steps but in one undo step.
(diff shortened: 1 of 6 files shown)


Reinmar:
The fix doesn't work for me. There is no tbody in the view when I remove the row.

jodator:
@Reinmar So after some digging up I ended up in fixing another set of tests (refs ckeditor/ckeditor5#6574). Apart of that I've also added removing empty table sections when down casting headings rows change.

jodator:
@Reinmar the current PR is probably the last one for row headings + removing rows problems. Any change of this code will blows something different and the longer explanation is here: ckeditor/ckeditor5#6630.

PR #304, opened 2020-04-14
Custom keyboard navigation through tables

niegowski:

Other: Custom keyboard navigation through tables. Closes ckeditor/ckeditor5#3267. Closes ckeditor/ckeditor5#3286.

---


oleq:
I think the fact that I can navigate across the lines downwards using the down arrow key but using the arrow up key moves the caret to the first position in the cell is a bug.

oleq:
☝️ I see that it works for multiple blocks, one after another, but in the screencast its a single block with wrapped content.

oleq:
In a case where the widget is the only child of the cell, there's a lockdown. Arrow keys won't let me navigate to neighboring cells.

image

src/tablenavigation.js
```diff
@@ -0,0 +1,480 @@
...
+		}
+
+		const viewRange = editing.mapper.toViewRange( modelRange );
+		const domRange = domConverter.viewRangeToDom( viewRange );
+		const rects = Rect.getDomRangeRects( domRange );
+
+		let boundaryVerticalPosition;
+
+		for ( let i = 0; i < rects.length; i++ ) {
+			const idx = isForward ? rects.length - i - 1 : i;
```
oleq, line 353: Can we move this to the `for()` statement? Now we have `idx` and `i` which do the same thing but the former is directional. Maybe let's just have different loop configurations (start, condition, incrementation/decrementation) for different `isForward` values.

src/tablenavigation.js
```diff
@@ -0,0 +1,480 @@
...
+			const rect = rects[ idx ];
+
+			// We need to check if current `rect` is container for following Rects.
+			const nextRect = rects.find( ( rect, i ) => i > idx && rect.width > 0 );
```
oleq, line 357: When does it happen? Some example would be great.

---

BTW: is this code OK in the first place? Shouldn't the `nextRect` be directional? Like when forward, then checking from `idx` onwards and when backward – in the opposite direction?

Anyway, I think this could be improved in terms of performance. For every `rect` of the main `for()` loop `find()`` will check dozens of `rects` before finding the `nextRect`, while it could be a loop that starts from `idx` instead.

src/tablenavigation.js
```diff
@@ -0,0 +1,480 @@
...
+		if ( column < 0 ) {
+			column = lastColumn;
+			row--;
+		} else if ( column > lastColumn ) {
+			column = 0;
+			row++;
+		}
+
+		const cellToSelect = tableMap.find( cellInfo => cellInfo.row == row && cellInfo.column == column ).cell;
+		const isForward = [ 'right', 'down' ].includes( direction );
```
oleq, line 446: KISS 😛

src/tablenavigation.js
```diff
@@ -0,0 +1,480 @@
...
+		} );
+	}
+}
+
+// Returns 'true' if provided key code represents one of the arrow keys.
+//
+// @private
+// @param {Number} keyCode
+// @returns {Boolean}
+function isArrowKeyCode( keyCode ) {
```
oleq, line 460: I think it's time we had an utils for it in `keyboard.js`. Use cases:

```
6 results - 3 files

⚙️ engine • src/view/observer/fakeselectionobserver.js:
   52: 			if ( selection.isFake && _isArrowKeyCode( data.keyCode ) && this.isEnabled ) {
  119: function _isArrowKeyCode( keyCode ) {

table • src/tablenavigation.js:
  165: 		if ( isArrowKeyCode( keyCode ) ) {
  460: function isArrowKeyCode( keyCode ) {

widget • src/widget.js:
  173: 		if ( isArrowKeyCode( keyCode ) ) {
  421: function isArrowKeyCode( keyCode ) {

```

(this could be a follow-up).

src/tablenavigation.js
```diff
@@ -0,0 +1,480 @@
...
+	 *
+	 * @private
+	 * @param {module:utils/eventinfo~EventInfo} eventInfo
+	 * @param {module:engine/view/observer/domeventdata~DomEventData} domEventData
+	 */
+	_handleTabOnSelectedTable( domEventData, cancel ) {
+		const editor = this.editor;
+		const selection = editor.model.document.selection;
+
+		if ( !selection.isCollapsed && selection.rangeCount === 1 && selection.getFirstRange().isFlat ) {
```
oleq, line 65: I think we can easily drop this `if()` because `Selection#getSelectedElement()` does it internally anyway.

niegowski: Methods `_handleTabOnSelectedTable` and `_getTabHandler` are just moved from `TableEditing` without any changes.

oleq:
This one also looks like a bug (arrow down should always navigate to the image and the cell below):

niegowski:
> This one also looks like a bug (arrow down should always navigate to the image and the cell below):
> 
> 

This should be fixed now in eebd6f2.

niegowski:
cc @ckeditor/qa-team Can you have a look at this? Navigation over table cells should be as intuitive as possible. Note that on the first/last text lines in a cell the caret is moved to first/last position in that cell (it's expected behavior). It's using DOM client rects so we should test it in different browsers.

Mgsy:
If a cell contains a blockquote, pressing <kbd>Arrow down</kbd> makes the selection jumps to the end of the blockquote, instead of a line below.

table

Mgsy:
We've tested it in all browsers and besides the issue above, it works fine 👍

niegowski:
> ## Firefox only
> If a cell contains a blockquote, pressing Arrow down makes the selection jumps to the end of the blockquote, instead of a line below.
> 
> table

This is how Firefox handles `overflow: hidden` for blockquote. It works that way even without tables. We use `overflow: hidden` there to trigger "Block formatting context" for blockquote to handle positioning next to floating widgets.

src/table.js
```diff
@@ -36,7 +37,7 @@ export default class Table extends Plugin {
 	 * @inheritDoc
 	 */
 	static get requires() {
-		return [ TableEditing, TableUI, TableSelection, TableClipboard, Widget ];
+		return [ TableEditing, TableUI, TableSelection, TableClipboard, TableNavigation, Widget ];
```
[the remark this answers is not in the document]
niegowski: Updated in 85b4dea.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+			const selectedElement = selection.getSelectedElement();
+
+			if ( !selectedElement || !selectedElement.is( 'table' ) ) {
+				return;
+			}
+
+			cancel();
+
+			editor.model.change( writer => {
+				writer.setSelection( writer.createRangeIn( selectedElement.getChild( 0 ).getChild( 0 ) ) );
```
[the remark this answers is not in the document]
niegowski: <kbd>Tab</kbd> handling implementation was only moved from `TableEditing` without modifications.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+		if ( selectedCells.length ) {
+			const tableCell = isForward ? selectedCells[ selectedCells.length - 1 ] : selectedCells[ 0 ];
+
+			this._navigateFromCellInDirection( tableCell, direction );
+
+			return true;
+		}
+
+		// Abort if we're not in a table.
+		const tableCell = findAncestor( 'tableCell', selection.focus );
```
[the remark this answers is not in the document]
niegowski: Updated in 85b4dea.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+				return null;
+			}
+
+			return new ModelRange( firstRangePosition, position );
+		}
+	}
+
+	/**
+	 * Basing on provided `boundaries` range, finds first/last (depending on `direction`) element
+	 * that can contain `$text` (according to schema) and is visible in the view.
```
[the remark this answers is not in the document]
niegowski: Updated.

src/tablenavigation.js
oleq: Created a follow-up https://github.com/ckeditor/ckeditor5/issues/6648.

tests/tablenavigation.js
```diff
@@ -0,0 +1,2277 @@
+import { getCode } from '@ckeditor/ckeditor5-utils/src/keyboard';
```
oleq, line 1: Missing license header.

tests/tablenavigation.js
```diff
@@ -0,0 +1,2277 @@
...
+import { getData as getModelData, setData as setModelData } from '@ckeditor/ckeditor5-engine/src/dev-utils/model';
```
oleq, line 2: We try to follow a convention in which named imports usually land after default imports. 

Also, we try (at least to some extend) to group imports (external packages in one group (image, horizontal line), in-package imports (table, table editing) in another, global helpers (`global`, `env`) in another).

This should probably be explained in https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/contributing/code-style.html (cc @Reinmar).

tests/tablenavigation.js
```diff
@@ -0,0 +1,2284 @@
...
+				editor.editing.view.document.fire( 'keydown', domEvtDataStub );
+
+				sinon.assert.calledOnce( domEvtDataStub.preventDefault );
+				sinon.assert.calledOnce( domEvtDataStub.stopPropagation );
+				assertEqualMarkup( getModelData( model ), modelTable( [
+					[ '11', '<blockQuote><paragraph>[foo]</paragraph></blockQuote>' ]
+				] ) );
+			} );
+
+			it( 'should listen with the lower priority than its children', () => {
```
oleq, line 221: I don't understand the description of this test. By reading the code I understand this is related to event bubbling and canceling but it's not clear what are the children and whose children they are. Maybe there's a way to re-phrase it?

niegowski: This is only moved from `TableEditing` tests. I wasn't even reading those.

tests/tablenavigation.js
```diff
@@ -0,0 +1,2284 @@
...
+
+						assertEqualMarkup( getModelData( model ), modelTable( [
+							[ '00', '01', '02' ],
+							[ '10', `${ paragraph }[${ hr }]`, '12' ],
+							[ '20', '21', '22' ]
+						] ) );
+					} );
+				} );
+
+				describe( 'contains image widget with caption and selection inside the caption', () => {
```
oleq, line 2068: What contains that image? TCs suggest this is about a cell where the initial selection is but it is not clear from the description. It could as well describe a place the selection should be anchored to after the navigation.

tests/tablenavigation.js
niegowski: I guess it's indentation handler.

tests/tablenavigation.js
niegowski: Table cell contains image.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+	 *
+	 * @private
+	 * @param {module:engine/model/element~Element} tableCell The table cell to start the selection navigation.
+	 * @param {'left'|'up'|'right'|'down'} direction Direction in which selection should move.
+	 */
+	_navigateFromCellInDirection( tableCell, direction ) {
+		const model = this.editor.model;
+
+		const table = findAncestor( 'table', tableCell );
+		const tableMap = [ ...new TableWalker( table, { includeSpanned: true } ) ];
```
jodator, line 436: Calculating this and other stuff in tables is prone to errors. Most likely when modifying a table but even those calculations could be unintentionally be wrong.

In other words, please use `TableUtils` for those:

1.  `const lastRow = tableUtils.getRows( table ) - 1`
2.  `const lastColumn = tableUtils.getColumns( table ) - 1`
3.  `const { row, column } = tableUtils.getCellLocation( tableCell )`

I'd feel safer to have this using `TableUtils`. Also using the above we don't have to iterate over spanned cells.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+				row += currentCellInfo.rowspan;
+				break;
+		}
+
+		const isOutsideVertically = row < 0 || row > lastRow;
+		const isBeforeFirstCell = column < 0 && row <= 0;
+		const isAfterLastCell = column > lastColumn && row >= lastRow;
+
+		// Note that if the last table cell is row-spanned then isAfterLastCell will never be true but we don't know
+		// if user was navigating on the last row or not, so let's allow him to stay in the table.
```
jodator, line 465: Do you have test for this? I think that this is not possible - we fix table if all cells have bigger rowspan than "visual" one.

Ie create a 2x4 table and merge all cells in the first column (it will have rowspan=4) then remove the last row. The rowspan will be reduced to 3.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+	 * @param {module:engine/model/selection~Selection} selection The current selection.
+	 * @param {Boolean} isForward The expected navigation direction.
+	 * @returns {module:engine/model/range~Range|null}
+	 */
+	_findTextRangeFromSelection( range, selection, isForward ) {
+		if ( isForward ) {
+			const position = selection.getLastPosition();
+			const lastRangePosition = this._getNearestVisibleTextPosition( range, 'backward' );
+
+			if ( !lastRangePosition || position.compareWith( lastRangePosition ) != 'before' ) {
```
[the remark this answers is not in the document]
jodator: I'd go with something like this:

```javascript
_findTextRangeFromSelection( range, selection, isForward ) {
	const model = this.editor.model;

	// We check outside the selection to the range boundary.
	const selectionEdge = isForward ? selection.getLastPosition() : selection.getFirstPosition();
	const checkStart = isForward ? selectionEdge : range.start;
	const checkEnd = isForward ? range.end : selectionEdge;

	// So the range to check is either
	// - from table cell start to first selection position
	// - or from last selection position to the end of a cell.
	const rangeToCheck = model.createRange( checkStart, checkEnd );

	const nearestVisibleTextPosition = this._getNearestVisibleTextPosition( rangeToCheck, isForward ? 'backward' : 'forward' );

	// The range must contain text - so it cannot be collapsed.
	if ( !nearestVisibleTextPosition || selectionEdge.isEqual( nearestVisibleTextPosition ) ) {
		return null;
	}

	const start = isForward ? selectionEdge : nearestVisibleTextPosition;
	const end = isForward ? nearestVisibleTextPosition : selectionEdge;

	return model.createRange( start, end );
}
```

But it is my personal preference (I don't like reading big ifs).

The point here is to:

1.  Don't use `Range` constructor.
2.  The range

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+// @private
+// @param {Number} keyCode
+// @param {Boolean} isLtrContent The content language direction.
+// @returns {'left'|'up'|'right'|'down'} Arrow direction.
+function getDirectionFromKeyCode( keyCode, isLtrContent ) {
+	switch ( keyCode ) {
+		case keyCodes.arrowleft: return isLtrContent ? 'left' : 'right';
+		case keyCodes.arrowright: return isLtrContent ? 'right' : 'left';
+		case keyCodes.arrowup: return 'up';
+		case keyCodes.arrowdown: return 'down';
```
jodator, line 516: Please format it so each `return` is in new line (code style consistency with existing code).

jodator:
And I have and inconsistency in the behaviour. Basically the feature works like the Google docs but - I don't find it OK but maybe it's just me.

So Google docs navigating with arrow (gets "trapped"):



So Google docs navigating with TAB (as in CKEditor - navigate on the same row):



Browsers (Chrome & Firefox) with arrows (similar to TAB action):



Similarly to browser, the Libre office follows the TAB behaviour pattern for arrows:



So as long we think that having different behaviour for TAB is OK (so as in GDocs) then it is a minor issue.

cc @oleq .

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+	 * Checks if {@link module:engine/model/element~Element element} placed next to the current
+	 * {@link module:engine/model/selection~Selection model selection} exists and is marked in
+	 * {@link module:engine/model/schema~Schema schema} as `object`.
+	 *
+	 * @private
+	 * @param {module:engine/model/selection~Selection} modelSelection The selection.
+	 * @param {Boolean} forward Direction of checking.
+	 * @returns {module:engine/model/element~Element|null}
+	 */
+	_getObjectElementNextToSelection( modelSelection, forward ) {
```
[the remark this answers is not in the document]
niegowski: This method is only to detect cases when `Widget` plugin should handle this. It's trying to replicate behavior from original method: [`_getObjectElementNextToSelection`](https://github.com/ckeditor/ckeditor5-widget/blob/62735277c6f270b47a0c2723644e889c7b335dff/src/widget.js#L326).

Because of [stop propagation](https://github.com/ckeditor/ckeditor5-widget/blob/62735277c6f270b47a0c2723644e889c7b335dff/src/widget.js#L250) of keydown events, `TableNavigation` needs to listen on higher priority than `Widget`'s `keydown` listener.

src/tablenavigation.js
niegowski: This comment was describing this situation:
Apr-22-2020 20-19-40
This is the same behavior as in GDocs.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+
+		const startPosition = direction == 'forward' ? boundaries.start : boundaries.end;
+
+		const treeWalker = new TreeWalker( { direction, boundaries, startPosition } );
+
+		for ( const { nextPosition, item } of treeWalker ) {
+			if ( schema.checkChild( nextPosition, '$text' ) ) {
+				const viewElement = mapper.toViewElement( item );
+
+				if ( viewElement && !viewElement.hasClass( 'ck-hidden' ) ) {
```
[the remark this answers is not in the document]
Reinmar: Not sure if you talked with @niegowski about this but we discussed this and I didn't see a different option that would be safe and simple. 

However, this bit definitely requires a very good comment on why it's needed.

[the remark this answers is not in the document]
jodator: @Reinmar yes that is why I've commented out the last part. Sorry to leave it uncertain :/ yes I don't see other option now.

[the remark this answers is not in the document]
jodator: I don't like but we don't have better :man\_shrugging:. We don't have to touch it.

src/tablenavigation.js
jodator: As @niegowski mentionend `table` map is reused also to navigate so it makes sense to leave it as-is.

src/tablenavigation.js
jodator:
```suggestion
		// Note that if the table cell at the end of a row is row-spanned then isAfterLastCell will never be true.
		// However, we don't know if user was navigating on the last row or not, so let's stay in the table.
```

src/tablenavigation.js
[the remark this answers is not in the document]
jodator: OK I didn't read the full history of this review. It is indeed copied from there.

Let's do a follow-up to extract it somewhere.

tests/tablenavigation.js
jodator: That's on me probably and it is a moved code. Let's leave it for now.

src/tablenavigation.js
```diff
@@ -0,0 +1,515 @@
...
+		const objectElement = selection.getSelectedElement();
+
+		if ( objectElement && model.schema.isObject( objectElement ) ) {
+			return;
+		}
+
+		// If next to the selection there is an object then this is not the cell boundary (widget handler should handle this).
+		if ( this._getObjectElementNextToSelection( selection, isForward ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
jodator: Misuse of review tools AFAICS by me.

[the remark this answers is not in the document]
jodator: Misuse of review tools AFAICS by me.

jodator:
The test that fails is from Firefox. I don't see what is causing this in this PR so I'm merging it and try fix CI later.

PR #306, opened 2020-04-20
I/6521: Add check for heading columns in merge cells command.

jodator:

Other: Merge cells command should not merge cells from heading columns with other cells. Closes ckeditor/ckeditor5#6521.

---


jodator:
I've found an undo bug: ckeditor/ckeditor5#6634. Also on master so a separate ticket.

src/commands/mergecellscommand.js
```diff
@@ -199,16 +199,45 @@ function getBiggestRectangleArea( rows, columns ) {
...
 
-	const firstCellIsInBody = rowIndexes.first > headingRows - 1;
-	const lastCellIsInBody = rowIndexes.last > headingRows - 1;
+	// Calculating row indexes is a bit cheaper so if this check fails we can't merge.
+	if ( !areIndexesInSameSection( rowIndexes, headingRows ) ) {
+		return false;
+	}
+
+	const headingColumns = parseInt( table.getAttribute( 'headingColumns' ) || 0 );
+	const columnIndexes = getColumnIndexes( tableCells );
```
niegowski, line 229: I know that this is not the scope of this PR but `getColumnIndexes` confused me a little, maybe it should be sth like `getFirstLastColumnIndexes`? Same for `getRowIndexes` -> `getFirstLastRowIndexes`.

jodator: > I know that this is not the scope of this PR

Yep, for me, it is outside a PR's scope. I'm fine with opening a new PR/issue for this. But here I didn't touch that code.

jodator:
> I would also add test like this:

Added this in 6c3ad1d  kinda redundant but why not.

